PR #11, opened 2023-05-18
Correct repo name in the enforce issue script

dkurepa:
https://github.com/dotnet/arcade/issues/13308

### DIFF
--- a/eng/enforce-issue.ps1
+++ b/eng/enforce-issue.ps1
@@ -1,12 +1,14 @@
 [CmdletBinding()]
 param (
 	[Parameter(Mandatory = $True)]
-	$PullRequestNumber
+	$PullRequestNumber,
+	[Parameter(Mandatory = $True)]
+	$RepositoryName
 )
 
 $prDetail = Invoke-WebRequest `
 	-UseBasicParsing `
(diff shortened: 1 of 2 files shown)


MilenaHristova (approved):

eng/enforce-issue.ps1
```diff
@@ -6,7 +6,7 @@ param (
 
 $prDetail = Invoke-WebRequest `
 	-UseBasicParsing `
-	-Uri "https://api.github.com/repos/dotnet/arcade-services/pulls/$PullRequestNumber" `
+	-Uri "https://api.github.com/repos/dotnet/dnceng/pulls/$PullRequestNumber" `
```
premun, line 9: I'd say this script should accept the repo the same way it accepts PR ID?

andriipatsula (dismissed):
LGTM

eng/enforce-issue.ps1
dkurepa: True, we could make it universal so we don't have to change this when we're creating new repos. There should be a pre defined AzDo variable that has the repo name

premun (approved):

andriipatsula (approved):

PR #13, opened 2023-05-18
Correct the yaml that's still metnioning arcade-services

dkurepa:
https://github.com/dotnet/arcade/issues/13611

### DIFF
--- a/eng/deploy.yaml
+++ b/eng/deploy.yaml
@@ -72,7 +72,7 @@ stages:
       inputs:
         serviceConnection: ${{ parameters.DotNetStatusEndpoint }}
         method: POST
-        urlSuffix: /arcade-services/$(Build.BuildNumber)/start
+        urlSuffix: /dnceng/$(Build.BuildNumber)/start
 
 - stage: deploy
   displayName: Deploy
@@ -199,7 +199,7 @@ stages:
       inputs:
         serviceConnection: ${{ parameters.DotNetStatusEndpoint }}
         method: POST
-        urlSuffix: /arcade-services/$(Build.BuildNumber)/end
+        urlSuffix: /dnceng/$(Build.BuildNumber)/end
   - job: updateMetrics
     displayName: Update Grafana Metrics
     dependsOn:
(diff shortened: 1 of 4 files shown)


eng/deploy.yaml
```diff
@@ -72,7 +72,7 @@ stages:
       inputs:
         serviceConnection: ${{ parameters.DotNetStatusEndpoint }}
         method: POST
-        urlSuffix: /arcade-services/$(Build.BuildNumber)/start
+        urlSuffix: /dnceng/$(Build.BuildNumber)/start
```
dkurepa, line 75: @garath do we need to make some changes in grafana to accommodate this change?

garath: Nope. That part of the path is used as input. 

https://github.com/dotnet/arcade-services/blob/74e37c2dcf5e326a780b40dbbda2e6708055abc6/src/DotNet.Status.Web/Controllers/DeploymentController.cs#L49-L100

dkurepa: I see, thanks!

src/DotNet.Status.Web/DotNet.Status.Web/TimelineIssueTriage.cs
```diff
@@ -21,7 +21,7 @@ namespace DotNet.Status.Web;
 
 public class TimelineIssueTriage : ITimelineIssueTriage
 {
-    private static readonly string _docLink = "[Documentation](https://github.com/dotnet/arcade-services/blob/main/docs/BuildFailuresIssueTriage.md)";
+    private static readonly string _docLink = "[Documentation](https://github.com/dotnet/dnceng/blob/main/docs/BuildFailuresIssueTriage.md)";
```
riarenas, line 24: Looks like this file is not in the repo

riarenas (commented):
Only questionable thing is the missing doc, do we want to bring it into the repo as part of the PR?

src/DotNet.Status.Web/DotNet.Status.Web/TimelineIssueTriage.cs
dkurepa: Ahh yes, my bad I didn't check this. I added the file

dkurepa:
@riarenas could you approve my PR if it looks good to you now please?

PR #25, opened 2023-05-24
Minor text clean up

missymessa:
https://github.com/dotnet/arcade/issues/13590


missymessa:
Just noticed that the RCA issue was opened in Arcade and not here... I need to fix that.

PR #109, opened 2023-06-02
Support partial matching of branch names in build monitor

premun:
We will need to match `release/*` and `internal/release/*` branches of a pipeline I want to add

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
@@ -128,9 +128,33 @@ public class AzurePipelinesController : ControllerBase
         }
     }
 
+    private static bool IsBranchMonitored(string[] patterns, string branch)
+    {
+        foreach (string pattern in patterns)
+        {
+            if (pattern.EndsWith("*"))
+            {
+                if (branch.StartsWith(pattern[..^1], StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
+                }
+            }
+            else
+            {
+                if (string.Equals(branch, pattern, StringComparison.OrdinalIgnoreCase))
+                {
+                    return true;
+                }
+            }
+        }
+
+        return false;
+    }
+
     private async Task ProcessBuildNotificationsAsync(IAzureDevOpsClient client, Build build)
     {
         const string fullBranchPrefix = "refs/heads/";
+
         foreach (var monitor in _options.Value.Monitor.Builds)
         {
             if (!string.Equals(build.Project.Name, monitor.Project, StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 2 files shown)


src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
```diff
@@ -128,9 +128,33 @@ private bool IsIgnoredStatus(Build build)
         }
     }
 
+    private static bool IsBranchMonitored(string[] patterns, string branch)
+    {
+        foreach (string pattern in patterns)
+        {
+            if (pattern.EndsWith("*"))
```
tkapin, line 135: Could you reduce the nesting here for better readability? For me, combining the first expression into `(pattern.EndsWith("*") && branch.StartsWith(pattern[..^1], StringComparison.OrdinalIgnoreCase)) and keeping the second one without the else would be the most readable way to express this logic.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
```diff
@@ -143,23 +167,21 @@ private async Task ProcessBuildNotificationsAsync(IAzureDevOpsClient client, Bui
                 continue;
             }
 
-            if (monitor.Branches.All(mb => !string.Equals($"{fullBranchPrefix}{mb}",
-                    build.SourceBranch,
-                    StringComparison.OrdinalIgnoreCase)))
+            string prettyBranch = build.SourceBranch;
```
tkapin, line 170: A nit: got the idea that prettyBranch is the branch name without prefix, but I'd suggest calling it sourceBranchName (similar to FullName and Name used commonly in .NET).

tkapin (dismissed):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
premun: @tkapin the only reduced nesting I can see here is adding a `continue` at the end of the `if` and not having an else.

I think it's not possible to merge the nested `if`s like you propose. What you propose would end up matching the name even if there is `*` inside.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
```diff
@@ -128,9 +128,33 @@ private bool IsIgnoredStatus(Build build)
         }
     }
 
+    private static bool IsBranchMonitored(string[] patterns, string branch)
+    {
+        foreach (string pattern in patterns)
```
andriipatsula, line 133: Who controls/creates the patterns? Do we have a comment/documentation/validation that says that pattern `release/*` is valid and `*/*` is not?

premun: We do, it's strictly internal and there is not a lot of documentation all around.

Please see for examples: https://github.com/dotnet/dnceng/pull/110/files

andriipatsula (approved):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
tkapin: Resolved offline.

tkapin (approved):

Issue #349, opened 2023-07-11
Missing bootstrap script in dotnet/dnceng


missymessa:
Copy over the bootstrap script that was used in arcade-services, get rid of stuff not needed in dnceng. 

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description

no release notes.

missymessa:
This was merged

PR #350, opened 2023-07-11
Including missing bootstrap script

missymessa:
https://github.com/dotnet/dnceng/issues/349

### DIFF
--- a/bootstrap.ps1
+++ b/bootstrap.ps1
@@ -9,4 +9,4 @@ Set-StrictMode -Version 2.0
 New-LocalGroup -Name "DncEngConfigurationUsers" -ErrorAction Continue
 Add-LocalGroupMember -Group "DncEngConfigurationUsers" -Member $(whoami) -ErrorAction Continue
 
-dotnet run --project .\src\Microsoft.DncEng.Configuration.Bootstrap\Microsoft.DncEng.Configuration.Bootstrap.csproj -- -r "https://vault.azure.net"-r "https://management.azure.com"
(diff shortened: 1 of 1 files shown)


bootstrap.ps1
```diff
@@ -0,0 +1,13 @@
...
+  exit 1
+}
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version 2.0
+
+New-LocalGroup -Name "DncEngConfigurationUsers" -ErrorAction Continue
+Add-LocalGroupMember -Group "DncEngConfigurationUsers" -Member $(whoami) -ErrorAction Continue
+
+dotnet tool restore
```
riarenas, line 12: This is attempting to restore the bootstra-dnceng-configuration dotnet tool, but this repo doesn't know that tool exists so this will porbably fail? 

If we want to use dotnet tool restore, we should add this tool to the global tools configuration for the repo, such as we do in Helix-service: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-service?path=/.config/dotnet-tools.json&version=GBmain&line=5&lineEnd=6&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

garath: The source for this tool is actually in this repo at `src/Microsoft.DncEng.Configuration.Bootstrap/`. Consider executing the project directly?

riarenas: The benefit of the manifest file is we always get a specific version restored. I don't think that benefit matters for this scenario though and executing the project seems fine.

garath (approved):

PR #373, opened 2023-07-17
Migrate Telemetry Service Staging to Helix cluster

dkurepa:
Migrates the Telemetry Service staging environment to use the Helix SF cluster

### DIFF
--- a/.vault-config/telemetryint.yaml
+++ b/.vault-config/telemetryint.yaml
@@ -2,7 +2,7 @@ storageLocation:
   type: azure-key-vault
   parameters:
     subscription: cab65fc3-d077-467d-931f-3932eabf36d3
-    name: maestroint
+    name: TelemetryServiceStaging
 
 references:
   helixkv:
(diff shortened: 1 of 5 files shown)


dkurepa:
The service has already been successfully deployed to the `dotnet-eng-helix-int-c1` cluster.
The KeyVault hasn't been populated yet. I will wait for a second opinion before populating it, specifically on which storage account should be used for the health table

garath:
> The KeyVault hasn't been populated yet. I will wait for a second opinion before populating it, specifically on which storage account should be used for the health table

Is there some question about how it should be handled?

dkurepa:
> Is there some question about how it should be handled?

I think mostly about where we want the KeyVaults. I created them in the same Resource Group as the cluster, since that seemed natural. We can change that if you folks want

garath:
> I created them in the same Resource Group as the cluster, since that seemed natural.

Ah, I have no problem with that. It seems like a fine choice.

dkurepa:
In that case I'm ready to merge this. The secrets are all in their vaults and should be ready to go. I'll also kill the instance running in the maestro-int cluster once this is merged

garath:
If you really want a Gold Star, you could switch these secrets to use Managed Identities :smile:

But I realize that's probably out of scope for this change.

garath (approved):

dkurepa:
> If you really want a Gold Star, you could switch these secrets to use Managed Identities 😄
> 
> But I realize that's probably out of scope for this change.

I'll give this a try tomorrow, shouldn't be that hard. Will merge this for now tho

PR #477, opened 2023-07-31
Migrate telemetry service

dkurepa:
https://github.com/dotnet/arcade-services/issues/2303


.vault-config/telemetryint.yaml
```diff
@@ -2,7 +2,7 @@ storageLocation:
   type: azure-key-vault
   parameters:
     subscription: cab65fc3-d077-467d-931f-3932eabf36d3
-    name: maestroint
+    name: TelemetryServiceStaging
```
tkapin, line 5: A nit - you are changing the name from all lowercase to Pascal case. Please make sure this isn't braking some naming conventions we might eventually have.

src/Telemetry/TelemetryApplication/PublishProfiles/Int.xml
```diff
@@ -22,8 +22,8 @@
                                     ServerCertThumbprint="0123456789012345678901234567890123456789" />
   -->
   <ClusterConnectionParameters
-    ResourceGroupName="maestro-int-cluster"
-    ClusterName="maestro-int"
+    ResourceGroupName="dotnet-eng-helix-int-cluster-1"
```
tkapin, line 25: Is the "-1" necessary? Are we expecting to have another clusters there?

tkapin (approved):

.vault-config/telemetryint.yaml
dkurepa: It's not. maestroint was the name of the vault we used to use, but we don't really have a standard for this. For example, we have HelixStaging, or EngKeyVault, etc..

src/Telemetry/TelemetryApplication/PublishProfiles/Int.xml
dkurepa: That's the name of the resource group, so it has to be there. I don't know if there were plans to have more than one cluster

Issue #740, opened 2023-08-30
ValidateSecrets failures in eng services repos


riarenas:
We are seeing auth errors during the secret validation stage of the official dnceng build
https://dev.azure.com/dnceng/internal/_build/results?buildId=2256113&view=logs&j=1b26fc61-3fb8-52f6-954f-c497793aab18&t=7406d6b8-afb6-55bb-9ee1-f2266ae71496

### Release Note Category
- [ ] Feature changes/additions 
- [X] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
Fixed authentication issues with the secret manager application.

riarenas:
This is the same symptoms as https://github.com/dotnet/dnceng/issues/723, but in this case the pipeline is already using the expected build pool that should be able to auth. I'll start investigating this.

riarenas:
The only difference I can see is the image we use for both pipelines. The failing build was using the [1es-windows-2019 image](https://github.com/dotnet/dnceng/blob/c4a490a89b38b3fc0c6ab9182af5436f95e346c8/eng/deploy.yaml#L21C6-L21C54), while our passing weeklies are using the [1es-windows-2022](https://github.com/dotnet/dnceng/blob/c4a490a89b38b3fc0c6ab9182af5436f95e346c8/azure-pipelines-weekly.yml#L19) image

This does seem like it would matter, so I'll try to hack up a test build that runs this job using the updated image to validate if this is something we should hotfix to unblock deployments.

riarenas:
This came back with a vengeance and seems to be affecting all of our repos now.

premun:
I am on FR and will work on this tomorrow too. One thought - the dnceng-shared is where secretmanager is, I'll try to run the validation there to see if it works. We've recently moved to new set of Azure auth libraries in the repo that get credentials from the env better so maybe that will help.

riarenas:
This is also breaking the weekly pipelines as well.

garath:
> the dnceng-shared is where secretmanager is, I'll try to run the validation there to see if it works

Secret Manager [is in dnceng](https://github.com/dotnet/dnceng/tree/main/src/SecretManager), not dnceng-shared.

riarenas:
The way this is set up:

[...]

PR #745, opened 2023-08-30
[Hotfix] Use the 1es-windows-2022 image instead of 2019 for secret validation

riarenas:
This should fix https://github.com/dotnet/dnceng/issues/740

I Validated that this fixes the stage by running this build https://dev.azure.com/dnceng/internal/_build/results?buildId=2256263&view=results which shows the different outcomes of running in the 2022 image versus the 2019 image.

(Main builds are also broken, but I'm focusing on the rollout right now, and I'll let this merge back to fix main)

### DIFF
--- a/eng/deploy.yaml
+++ b/eng/deploy.yaml
@@ -22,7 +22,7 @@ stages:
   - job: ValidateSecrets
     pool: 
       name: NetCore1ESPool-Internal-NoMSI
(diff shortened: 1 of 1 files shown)


riarenas:
(I don't have a root cause as of why this fixes things. If someone thinks it's important to get to the bottom of this one, we can RCA the issue for someone to dig)

PR #758, opened 2023-09-01
Fix SecretManager authentication problems via Azure CLI

premun:
This is a fix to unblock pipelines getting hit by https://github.com/dotnet/dnceng/issues/740. It is a successor of https://github.com/dotnet/dnceng/pull/732.

It moves this repo onto a new set of Azure auth libraries since the current ones are deprecated and CG alerts about them already. They also fix the problem we have by getting credentials from new versions of Azure CLI that has been rolled out to new agent images.

This doesn't not clear the alert fully as the offending libraries are used in few more places but that is not the goal of this PR.
There will be one more follow-up that will probably move the extension helpers methods into `dnceng-shared` but this will get us going first.


premun:
@riarenas I might not be around to merge this but feel free to push into this PR and merge this to unblock pipelines.

It should be enough to flow dnceng (SecretManager) to other repos once this gets merged. The main build of dnceng should also go green after this.

riarenas (approved):

riarenas:
Thanks @premun! 

I'll try to flow this new version everywhere today and unblock our other pipelines.

Issue #760, opened 2023-09-01
Build monitor is not opening issues for dotnet/dnceng builds


riarenas:
There have been a few builds that should have opened issues in dnceng:
dnceng-ci: https://dev.azure.com/dnceng/internal/_build/results?buildId=2256910&view=results and https://dev.azure.com/dnceng/internal/_build/results?buildId=2256910&view=results

dnceng-weekly: https://dev.azure.com/dnceng/internal/_build/results?buildId=2256969&view=results and https://dev.azure.com/dnceng/internal/_build/results?buildId=2257888&view=results

As an example for build 2257888, The service hook reports as successful for delivery, with the event marking the build as failed, but no issue was opened. This means the web service received the request successfully, but was unable to open an issue. There might be information int the app insights traces.

![image](https://github.com/dotnet/dnceng/assets/23242101/4333ee5e-a9f5-4214-bfe2-380538efb6f6)

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description
No release note needed here.

missymessa:
Do you know, off hand, which service is supposed to be handling this? Is it the WorkItemProcessor in Helix Services?

riarenas:
Nope, this code lives in this repo.

The service hook is configured to route things to https://dotneteng-status.azurewebsites.net/api/azp/build-complete

https://github.com/dotnet/dnceng/blob/main/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs#L55

riarenas:
the logs don't show anything suspicious, and build monitor has been opening issues for other repos. I'll see if there's a misconfiguration for dnceng.

riarenas:
There's indeed a typo: https://github.com/dotnet/dnceng/blob/main/src/DotNet.Status.Web/DotNet.Status.Web/.config/settings.json#L126 dnc**neg**

riarenas:
This was rolled out on 09/20

PR #766, opened 2023-09-01
fix typo in pipeline path for dnceng

riarenas:
Fixes https://github.com/dotnet/dnceng/issues/760. This should get us issues opened when there are failures in dnceng builds


missymessa (approved):
I'm sorry :(

Issue #765, opened 2023-09-01
dnceng-official broken


riarenas:
The build is broken but build monitor is not working for this repo https://github.com/dotnet/dnceng/issues/760. Investigate and fix.

https://dnceng.visualstudio.com/internal/_build/results?buildId=2257959&view=results

### Release Note Category
- [ ] Feature changes/additions 
- [x] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
Fixed regressions in the CI pipeline caused by auth library changes

riarenas:
This is most likely a consequence of the new auth libraries, and will probably need some reaction to the code in both the Grafana publishing and the scenario tests.

riarenas:
This is a side effect of https://github.com/dotnet/dnceng/pull/758.

* for the secret-manager scenario tests, we don't initialize the azure CLI, which is causing failures with the new flow of trying to obtain the cli credentials first.
* For the grafana publishing. It looks like we need to update our authentication code for the new libraries. 

We could revert #758, but that would mean making secret-manager break during synchronization and validation runs. I think it's less impactful if we leave this repo broken until we fix things up.

premun:
Sorry, totally missed this.

> I think it's less impactful if we leave this repo broken until we fix things up.

I agree.

Having a quick glance - we're probably failing to connect to a keyvault here: https://github.com/dotnet/dnceng/blob/bf68694c553f692e4f8b50883f79ffcd5ec64e5e/src/Monitoring/Sdk/DeployPublisher.cs#L284

premun:
This bit is also interesting:

```
error MSB4018: Azure.Identity.AuthenticationFailedException: ClientSecretCredential authentication failed: The type initializer for 'Microsoft.Identity.Client.Platforms.net6.MsalJsonSerializerContext' threw an exception. [D:\a\_work\1\s\src\Monitoring\Monitoring.ArcadeServices\Monitoring.ArcadeServices.proj]
error MSB4018:  ---> System.TypeInitializationException: The type initializer for 'Microsoft.Identity.Client.Platforms.net6.MsalJsonSerializerContext' threw an exception. [D:\a\_work\1\s\src\Monitoring\Monitoring.ArcadeServices\Monitoring.ArcadeServices.proj]
[...]

PR #789, opened 2023-09-08
Fix auth in Grafana dashboard publishing

premun:
Tested locally and it works. The key bit was the missing `JsonEncode` method so we're bringing the Json.Text library in explicitly

https://github.com/dotnet/dnceng/issues/765

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -88,7 +88,7 @@
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta1.21308.1" />
     <PackageVersion Include="System.Net.Http" Version="4.3.4" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="7.0.0" />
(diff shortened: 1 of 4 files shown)


tkapin (commented):
The code change looks good, but the CI throws the following error and complains about using System.Text.Json 7.0.0 when System.Text.Json (>= 7.0.1) is required.
```
src\DotNet.Status.Web\DotNet.Status.Web\DotNet.Status.Web.csproj(0,0): error NU1109: (NETCORE_ENGINEERING_TELEMETRY=Restore) Detected package downgrade: System.Text.Json from 7.0.1 to centrally defined 7.0.0. Update the centrally managed package version to a higher version. 
 DotNet.Status.Web -> Microsoft.Azure.Kusto.Ingest 11.3.1 -> System.Text.Json (>= 7.0.1) 
 DotNet.Status.Web -> System.Text.Json (>= 7.0.0)
```

tkapin (approved):

PR #820, opened 2023-09-15
Fixing Secret Manager tests in deployment

missymessa:
Fixes: https://github.com/dotnet/dnceng/issues/765


garath (approved):

eng/deploy.yaml
```diff
@@ -205,10 +205,10 @@ stages:
   pool:
     name: NetCore1ESPool-Internal
     demands: ImageOverride -equals 1es-windows-2019
-  dependsOn:
```
AlitzelMendez, line 208: is this no longer necessary?

I know we already merged this, I am just curious if this change is needed.

I was looking at the builds and before this change this was the pipeline 
image


and now it looks this way:
image

now is not running in parallel anymore, is this intended? do we need that the post deployment happens first?

Issue #885, opened 2023-09-26
[dotnet/dnceng] Replace `Microsoft.Azure.Services.AppAuthentication` with `Azure.Identity`


missymessa:
Split from https://github.com/dotnet/dnceng/issues/630

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description
No release notes necessary

missymessa:
Rolled out, closing.

PR #906, opened 2023-09-27
Refactor references of Microsoft.Azure.Services.AppAuthentication to Azure.Identity

missymessa:
https://github.com/dotnet/dnceng/issues/885

### DIFF
--- a/src/Microsoft.DncEng.DeployServiceFabricCluster/Helpers.cs
+++ b/src/Microsoft.DncEng.DeployServiceFabricCluster/Helpers.cs
@@ -2,11 +2,12 @@ using System.Net.Http.Headers;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Identity;
 using Microsoft.Azure.Management.Fluent;
 using Microsoft.Azure.Management.ResourceManager.Fluent;
 using Microsoft.Azure.Management.ResourceManager.Fluent.Authentication;
 using Microsoft.Azure.Management.ResourceManager.Fluent.Core;
-using Microsoft.Azure.Services.AppAuthentication;
 using Microsoft.Rest;
 using Microsoft.Rest.TransientFaultHandling;
 
@@ -15,20 +16,24 @@ namespace Microsoft.DncEng.DeployServiceFabricCluster;
 internal static class Helpers
 {
     private const string MsftAdTenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47";
-    private static readonly AzureServiceTokenProvider TokenProvider = new AzureServiceTokenProvider();
+    private static readonly DefaultAzureCredential TokenProvider = new DefaultAzureCredential();
 
     private class AzureCredentialsTokenProvider : ITokenProvider
     {
-        private readonly AzureServiceTokenProvider _inner;
+        private readonly DefaultAzureCredential _inner;
 
-        public AzureCredentialsTokenProvider(AzureServiceTokenProvider inner)
+        public AzureCredentialsTokenProvider(DefaultAzureCredential inner)
         {
             _inner = inner;
         }
 
         public async Task<AuthenticationHeaderValue> GetAuthenticationHeaderAsync(CancellationToken cancellationToken)
         {
-            string token = await _inner.GetAccessTokenAsync("https://management.azure.com", MsftAdTenantId);
+            AccessToken accessToken = await _inner.GetTokenAsync(
+                new TokenRequestContext(scopes: new string[] { "https://management.azure.com/.default" }, tenantId: MsftAdTenantId) { }
+            );
+
+            string token = accessToken.Token;
             return new AuthenticationHeaderValue("Bearer", token);
         }
     }
--- a/src/Microsoft.DncEng.DeployServiceFabricCluster/Microsoft.DncEng.DeployServiceFabricCluster.csproj
+++ b/src/Microsoft.DncEng.DeployServiceFabricCluster/Microsoft.DncEng.DeployServiceFabricCluster.csproj
@@ -10,9 +10,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+	  <PackageReference Include="Azure.Identity" />
     <PackageReference Include="Azure.Storage.Blobs" />
     <PackageReference Include="Microsoft.Azure.Management.Fluent" />
-    <PackageReference Include="Microsoft.Azure.Services.AppAuthentication" />
     <PackageReference Include="Microsoft.Extensions.Logging.Console" />
     <PackageReference Include="Mono.Options" />
     <PackageReference Include="Newtonsoft.Json" />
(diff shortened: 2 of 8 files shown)


src/Microsoft.DncEng.DeployServiceFabricCluster/Microsoft.DncEng.DeployServiceFabricCluster.csproj
```diff
@@ -10,9 +10,9 @@
   </PropertyGroup>
 
   <ItemGroup>
+	  <PackageReference Include="Azure.Identity" />
```
garath, line 13:
```suggestion
    <PackageReference Include="Azure.Identity" />
```
Spacing is wonky.

src/RolloutScorer/RolloutScorerAzureFunction/RolloutScorerFunction.cs
```diff
@@ -1,6 +1,5 @@
 using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
-using Microsoft.Azure.Services.AppAuthentication;
```
garath, line 3: Oh, just an unnecessary using? Nice!

missymessa: Yeah, there were a couple of those I was able to clean up :)

garath (commented):
For RolloutScorere, this bit exists in deploy.yaml. The environment variable `AzureServicesAuthConnectionString` is used by `AzureServiceTokenProvider`. Ensure this value is not actually used and, if it is, reformat it in a way acceptable to `EnvironmentTokenCredential`.

https://github.com/dotnet/dnceng/blob/c55ff7197477466713293b2c40fc138023216ed2/eng/deploy.yaml#L142-L161

src/Microsoft.DncEng.DeployServiceFabricCluster/Helpers.cs
```diff
@@ -15,20 +16,24 @@ namespace Microsoft.DncEng.DeployServiceFabricCluster;
 internal static class Helpers
 {
     private const string MsftAdTenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47";
-    private static readonly AzureServiceTokenProvider TokenProvider = new AzureServiceTokenProvider();
+    private static readonly DefaultAzureCredential TokenProvider = new DefaultAzureCredential();
```
garath, line 19: This is the scary bit where the behavior differences between AzureServiceTokenProvider and DefaultAzureCredential will pop up. All downstream consumers will need to be checked for these differences (for example, if credentials are passed by environment variable).

missymessa: While refactoring, all references to the changed code was within this file. If I'm missing something, let me know.

garath: This looks fine, I'm just highlighting that this is where behavior differences are so anything that consumes this code will have to also be changed accordingly.

missymessa:
> For RolloutScorere, this bit exists in deploy.yaml. The environment variable `AzureServicesAuthConnectionString` is used by `AzureServiceTokenProvider`. Ensure this value is not actually used and, if it is, reformat it in a way acceptable to `EnvironmentTokenCredential`.
> 
> https://github.com/dotnet/dnceng/blob/c55ff7197477466713293b2c40fc138023216ed2/eng/deploy.yaml#L142-L161

I couldn't find anywhere the `AzureServicesAuthConnectionString` was used, but the other four variables were being used, so I deleted both of them. We'll see if that breaks things when it comes time to deploy RolloutScorer.

garath:
> I couldn't find anywhere the AzureServicesAuthConnectionString was used, but the other four variables were being used, so I deleted both of them. We'll see if that breaks things when it comes time to deploy RolloutScorer.

To be clear, AzureServicesAuthConnectionString will not be referenced in our code anywhere. It is consumed by the Microsoft.Azure.Services.AppAuthentication library. So a simple string search will return nothing. The only way to be sure is to grok how the project is expecting authentication to work.

missymessa:
> > I couldn't find anywhere the AzureServicesAuthConnectionString was used, but the other four variables were being used, so I deleted both of them. We'll see if that breaks things when it comes time to deploy RolloutScorer.
> 
> To be clear, AzureServicesAuthConnectionString will not be referenced in our code anywhere. It is consumed by the Microsoft.Azure.Services.AppAuthentication library. So a simple string search will return nothing. The only way to be sure is to grok how the project is expecting authentication to work.

I'm going to assume that since RolloutScorer doesn't make use of the references to the `Microsoft.Azure.Services.AppAuthentication` library that was removed, that it doesn't use that parameter in the project deployment. (RolloutScorer looks like it was already updated to use `Azure.Identity`, too). 

Unless there's something automagical going on in the background for how Azure Functions are deployed?

garath:
> To be clear, AzureServicesAuthConnectionString will not be referenced in our code anywhere. It is consumed by the Microsoft.Azure.Services.AppAuthentication library.

Oh, one idea: Write a new [TokenCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential?view=azure-dotnet) class that reads that environment variable and adapts it into something the new library understands.

garath:
> I'm going to assume that since RolloutScorer doesn't make use of the references to the Microsoft.Azure.Services.AppAuthentication library that was removed, that it doesn't use that parameter in the project deployment. (RolloutScorer looks like it was already updated to use Azure.Identity, too).

Having now looked through the RolloutScorer code, I agree, that appears to be the case.

missymessa:
> > To be clear, AzureServicesAuthConnectionString will not be referenced in our code anywhere. It is consumed by the Microsoft.Azure.Services.AppAuthentication library.
> 
> Oh, one idea: Write a new [TokenCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential?view=azure-dotnet) class that reads that environment variable and adapts it into something the new library understands.

I don't think this will be necessary. 

The migration guidance doc calls out how to migrate those connection strings near the beginning: https://learn.microsoft.com/en-us/dotnet/api/overview/azure/app-auth-migration?view=azure-dotnet 

Given that the migration to use `Azure.Identity` in RolloutScorer was back in [Dec 2022](https://github.com/dotnet/dnceng/commit/f343f1abf091879c5bef05a41f870668e3b771c8), and we've been able to deploy it without issues since then, it doesn't look like that variable is necessary. 

Tagging @dkurepa in case he has further insight, since he did the original migration.

garath (approved):

garath (approved):

garath:
> Given that the migration to use Azure.Identity in RolloutScorer was back in Dec 2022, and we've been able to deploy it without issues since then, it doesn't look like that variable is necessary.

Looking through the history I suspect the change in dotnet/arcade-services#2150 did (unintentionally?) change how authentication worked. 

But that's the past.

missymessa:
Going to merge this. @dkurepa let me know if there's anything else that should've been changed when I removed the AppAuth references.

Issue #924, opened 2023-09-29
Fix scenario in which issues were still open with label First Responder instead of Ops - First Responder


AlitzelMendez:
Example of the issue happening:
- https://github.com/dotnet/dnceng/issues/902
- https://github.com/dotnet/dnceng/issues/899
- https://github.com/dotnet/dnceng/issues/918

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
No release note needed

PR #925, opened 2023-09-29
Fix scenario in which issues were still open with label First Respond…

AlitzelMendez:
Issue: https://github.com/dotnet/dnceng/issues/924


ilyas1974 (approved):
thank you for doing this.

PR #936, opened 2023-10-02
Move team documentation from arcade

premun:


missymessa (approved):
I'd love to get some of this cleaned up, but that work is outside the scope of the work of just migrating it out of Arcade.

premun:
Yeah, I think many of those designs were turned into real projects that got their real documentation written too

Issue #1148, opened 2023-10-24
Refine #961 work slightly


dougbu:
#961 added some cleanup to the setup.sh script used in the [`grafana-setup`](https://dev.azure.com/dnceng/internal/_build?definitionId=1313) pipeline. Unfortunately, the added commands could fail when the target machine is in an extremely low disk space situation. Aim of this issue is to do all cleanup before trying to (for example) download new files.

This work is a continuation of what we did in #961 and a small step toward fixing #967. Work includes one item from #967 but not a rewrite to avoid the significant 4K logging limit mentioned there. (That logging limit should be considered the core priority within #967.)

### Release Note Category

- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements

### Release Note Description

Refined the scripts used in an internal pipeline we use for a service deployment.

dougbu:
should already have closed this 😦

Issue #967, opened 2023-10-07
Suggest updating grafana-setup.yml in dnceng


dougbu:
grafana-setup.yml does a few odd things, including relying on variables set in the UI instead of top-level parameters and
``` yaml
trigger:
  branches:
    exclude:
      - '*'
```
instead of `trigger: none`.

also seems best practice to include `pr: none` as well, especially if we ever do internal PRs b/c they'd trigger the pipeline by default.

on the first point, top-level parameters would make it much easier to submit manual runs (which seems to be the only runs we want). in addition, would be good for the YAML to figure out `GrafanaDomain` and `TargetVmName` based on `GrafanaEnvironment`. one top-level parameter to rule them all :grin:

### Release Note Category

- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements

### Release Note Description

Cleaned up the grafana-setup pipeline used for internal deployment of one of our services.

dougbu:
In addition, the 4K console logging limit is problematic. Might be better to use the `SSH@0` task w/ a SSH service connection.

Problematic why you say❔ It means successful runs don't include early command output and debugging commands like `df` and `du` bits in setup.sh aren't providing any information. Of course, failing or hanging commands exit the `inlineScript` early and that'll increase the likelihood of seeing the full output.

ilyas1974:
As the priority of this has not increased in over a year, we do not have the resources to address this currently. Should we be in error, please reopen the issue

PR #1149, opened 2023-10-24
Clean up setup-grafana cleanup slightly

dougbu:
- see https://github.com/dotnet/dnceng/issues/1148
- do cleanup before initial download in setup.sh
- just `apt-get check` instead of `apt-get install --fix-broken`
  - fixing broken packages is not commonly necessary
- also display /tmp and /var disk usage
- nit: don't use `findstr` in a PowerShell script
  - use `pscore` b/c that supports the `||` operator
  - a small part of https://github.com/dotnet/dnceng/issues/967

### DIFF
--- a/src/Monitoring/grafana-init/setup.sh
+++ b/src/Monitoring/grafana-init/setup.sh
@@ -96,11 +96,6 @@ fi
 
 export DEBIAN_FRONTEND=noninteractive
 
-# This is the grafana package repo that allos us to apt-get grafana
-# If we don't trust grafana.com, we're in hot water already, so this is fine
-wget -q -O - https://packages.grafana.com/gpg.key | sudo apt-key add -
-add-apt-repository "deb https://packages.grafana.com/oss/deb stable main"
-
 # Before
 df --human-readable --inodes
 df --human-readable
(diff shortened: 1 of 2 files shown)


dougbu:
I ran a slightly older version of this branch (prior to rebasing on today's main) to deploy Grafana to staging in [#20231006.4](https://dev.azure.com/dnceng/internal/_build/results?buildId=2285643&view=results) and production in [#20231006.5](https://dev.azure.com/dnceng/internal/_build/results?buildId=2285668&view=results). Both builds and their deployments were successful.

Both builds also hide the `df` and `du` command output (added in #961 and here) because that comes before the 4K chunk that `az vm run-command invoke` deigns to output ☹️

missymessa (approved):
LGTM shipit

src/Monitoring/grafana-init/setup.sh
```diff
@@ -109,11 +104,17 @@ df --human-readable
 apt-get autoremove --yes
 apt-get autoclean --yes
 apt-get clean --yes
-apt-get install --fix-broken --yes
+apt-get check --yes
```
garath, line 107: Anything to be concerned about with the difference between `install` and `check`?

garath (approved):

Issue #1368, opened 2023-11-06
Weekly build failures are being tagged with incorrect labels


riarenas:
The issues for weekly build failures should be routed to First responders.

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description
Added first responders label to weekly build failures

riarenas:
PR: https://github.com/dotnet/dnceng/pull/1370

riarenas:
This was deployed yesterday.

PR #1369, opened 2023-11-06
fix build monitor configuration for weekly builds

riarenas:
https://github.com/dotnet/dnceng/issues/1368

This makes it so weekly builds get tagged for First responders to fix. It also removes the FRops sections, as we no longer use that label.


riarenas:
OOF nevermind, my fork is super out of date. I'll fix this up and reopen with a cleaner PR

PR #1522, opened 2023-12-06
Adding Vault to track Sevice Connection Expirations

ilyas1974:
going to use secret manager to track service connects that are expiring.


garath:
I propose creating a new simple secret type to help manage service connections in #1526. I think these two PRs should become friends.

ilyas1974:
Abandoning this PR as @garath made some changes to the secret type. Will redo this in the new format

PR #1526, opened 2023-12-07
Add simple service-connection type for Secret manager

garath:
Resolves [dotnet/dnceng#1527](https://github.com/dotnet/dnceng/issues/1527)

This adds a new secret type to secret-manager to assist in tracking expiration of secrets used in service connections. 

To do so, it creates a new entry in the specified key vault, just like any secret-manager thing, representing the service connection and its secret. The actual secret is not stored as secret-manager cannot do anything to help rotation except provide guidance. Instead, the entry provides an inventory of service connections and uses the metadata to track expirations in a way consistent with other secret management operations.

This is not intended to solve all problems with secrets and service connections. Its primary goal is to replace the "schedule a meeting on the expiration day" method used today with something more aligned with team values and processes (which, to me, is best done with the secret-manager tool). I'm sure we will make incremental improvements as time allows.

This is meant to make @ilyas1974's life easier as he is driving the effort to make service connections less painful.

Some implementation notes:

- The actual secret is not stored as secret-manager can't actually do anything to help except provide some links. The tool will quietly insert an empty string for the secret value.
- The `name` and `description` field aren't actually used in the tool, but they make working with service connections much easier so I felt an explicit spot was acceptable
- The `id`, `organization` and `project` are the three elements that uniquely identify a service connection. Aside from making the reference unambiguous, they are used by the tool to display a link going directly to the service connection configuration in the Azure DevOps portal.
- It is intended that an internal wiki page be created to document any other important details in this wide-ranging topic.

### DIFF
--- a/src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceConnectionSecretType.cs
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceConnectionSecretType.cs
@@ -16,7 +16,7 @@ public class ServiceConnectionSecretType : SecretType<ServiceConnectionSecretTyp
         Service Connections have no standard way to rotate their tokens. Please follow instructions 
         in the DNCEng Internal Wiki at:
         
-        > https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1157/Generic-Service-Connections.
+        > https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1157
 
         Direct link to this service connection in the portal: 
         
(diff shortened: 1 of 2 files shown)


riarenas (approved):

riarenas (approved):

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceConnectionSecretType.cs
```diff
@@ -0,0 +1,64 @@
...
+[Name("service-connection")]
+public class ServiceConnectionSecretType : SecretType<ServiceConnectionSecretType.Parameters>
+{
+    private readonly IConsole _console;
+
+    private const string _helpMessage = """
+        Service Connections have no standard way to rotate their tokens. Please follow instructions 
+        in the DNCEng Internal Wiki at:
+        
+        > https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1157/Generic-Service-Connections.
```
ilyas1974, line 19:
```suggestion
        > https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1157
```

ilyas1974 (approved):
One change to the URL for the wiki.

AlitzelMendez (approved):

PR #2115, opened 2024-03-01
Prioritize AzureCliCredential for secret-manager

garath:
Resolves [dotnet/dnceng#2084](https://github.com/dotnet/dnceng/issues/2084)

Prioritize using AzureCliCredential to eliminate confusion in CI.

### DIFF
--- a/src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
@@ -1,0 +1,9 @@
+using System.Threading.Tasks;
+using Azure.Core;
+
+namespace Microsoft.DncEng.SecretManager;
+
+public interface ITokenCredentialProvider
+{
+    public Task<TokenCredential> GetCredentialAsync();
+}
--- a/src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
@@ -1,0 +1,22 @@
+using System;
+using System.Threading.Tasks;
+using Azure.Core;
+using Azure.Identity;
+
+namespace Microsoft.DncEng.SecretManager;
+
+public sealed class SecretManagerCredentialProvider : ITokenCredentialProvider
+{
+    // Expect AzureCliCredential for CI and local dev environments. 
+    // Use InteractiveBrowserCredential as a fallback for local dev environments.
+    private readonly Lazy<TokenCredential> _credential = new(() =>
+        new ChainedTokenCredential(
+            new AzureCliCredential(new AzureCliCredentialOptions { TenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47" }),
+            new InteractiveBrowserCredential(new InteractiveBrowserCredentialOptions() { TenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47" })
+        ));
+
+    public Task<TokenCredential> GetCredentialAsync()
+    {
+        return Task.FromResult(_credential.Value);
+    }
+}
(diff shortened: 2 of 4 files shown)


src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
garath, line 1: Creating a new interface for tokens undermines `Microsoft.DncEng.CommandLineLib` a bit, but unfortunately it does not quite let consumers override the authentication path. I'll take a quick look to see if there are reasonable changes that can be made. This change should still be taken first to unblock everything; it can take any future changes from CommandLineLib later.

garath:
I'm running the weekly build against this branch to confirm it works: [#2024022911 • Prioritize AzureCliCredential for secret-manager](https://dev.azure.com/dnceng/internal/_build/results?buildId=2391180&view=results)

src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
garath: Maybe something like dotnet/dnceng-shared#52

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.DncEng.SecretManager;
+
+public sealed class SecretManagerCredentialProvider : ITokenCredentialProvider
+{
+    // Expect AzureCliCredential for CI and local dev environments. 
+    // Use InteractiveBrowserCredential as a fallback for local dev environments.
+    private readonly Lazy<TokenCredential> _credential = new(() =>
+        new ChainedTokenCredential(
+            new AzureCliCredential(new AzureCliCredentialOptions { TenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47" }),
```
dkurepa, line 14: nit: we can put the tenantId in a const, or even use an existing const from somewhere else, maybe https://github.com/dotnet/dnceng-shared/blob/13ce0138bcbe745d6dc5a2d47a11a9a1980f8172/src/Microsoft.DncEng.Configuration.Extensions/ConfigurationConstants.cs#L13, since we already include include that package

src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
```diff
@@ -0,0 +1,9 @@
+using System.Threading.Tasks;
+using Azure.Core;
+
+namespace Microsoft.DncEng.SecretManager;
+
+public interface ITokenCredentialProvider
+{
+    public Task<TokenCredential> GetCredentialAsync();
```
premun, line 8:
```suggestion
    Task<TokenCredential> GetCredentialAsync();
```

premun (approved):

src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
garath: Why?

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
garath: Ugh, yes I made that change locally but forgot to push. Fixed in d655bb8c.

AlitzelMendez (approved):

src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
premun: Interfaces don't need these

PR #2309, opened 2024-03-20
Add Maestro Exceptions panel to grafana

dkurepa:
https://github.com/dotnet/arcade-services/issues/3427

### DIFF
--- a/src/Monitoring/Monitoring.ArcadeServices/dashboard/arcade-services/arcadeAvailability.dashboard.json
+++ b/src/Monitoring/Monitoring.ArcadeServices/dashboard/arcade-services/arcadeAvailability.dashboard.json
@@ -1828,7 +1828,9 @@
     },
     {
       "alert": {
-        "alertRuleTags": {},
+        "alertRuleTags": {
+          "NotificationId": "yek1lyodx21u1kzlpa62iz7mws0wcwpt"
+        },
         "conditions": [
           {
             "evaluator": {
@@ -1901,7 +1903,7 @@
               "mode": "none"
             },
             "thresholdsStyle": {
-              "mode": "off"
+              "mode": "line"
             }
           },
           "mappings": [],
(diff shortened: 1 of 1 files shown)


src/Monitoring/Monitoring.ArcadeServices/dashboard/arcade-services/arcadeAvailability.dashboard.json
```diff
@@ -1825,6 +1825,160 @@
       ],
       "title": "Helix AutoScaler Service Traces",
       "type": "timeseries"
+    },
+    {
+      "alert": {
+        "alertRuleTags": {},
```
dkurepa, line 1831: Looking at other alerts, I'm missing an `NotificationId` tag. Not sure how those are generated tho, is it just a random guid?

tkapin: Check with @garath

premun: It's a random guid I manually add when I create these in the UI. Not sure that is the right approach but definitely works.

premun:
@dkurepa there's no alert added for this panel?

Can we add the panel to the prod dashboard so that we can see how it would react to the last month's rise in exceptions?

dkurepa:
> @dkurepa there's no alert added for this panel?
> 
> Can we add the panel to the prod dashboard so that we can see how it would react to the last month's rise in exceptions?

I redid the entire panel and force pushed to my branch. You can see the panel I created and exported here https://dotnet-eng-grafana.westus2.cloudapp.azure.com/d/arcadeAvailability/service-availability?orgId=1&from=now-30d&to=now&viewPanel=46

premun:
Can you regenerate it again? I added a threshold so it's obvious when it will trigger

dkurepa:
> threshold

Sure, done

premun (approved):

PR #2772, opened 2024-05-06
Ol' Maestro, the deprecationing

mmitche:


missymessa:
Did you write this, or chatGPT? :D

Documentation/OnePagers/ol-maestro-deprecation.md
```diff
@@ -0,0 +1,83 @@
...
+- Should use GH and AzDO APIs to determine the state of repos on either side of the mirror to determine whether repo changes need to happen.
+- Should use a MI to do mirror pushes to AzDO and fetches from AZDO.
+- Should report changes as well as failures to mirror.
+- Should not abort if a single mirror operation fails (complete and report at end).
+- Should operate in parallel with a throttling mechanism.
+- When the current pipeline completes, a new one should launch.
+
+### Component: Inter-branch Merge GitHub Action
+
+A GitHub Action should be developed that replaces the inter-branch merge functionality. It should have the following qualities:
```
chcosta, line 54: Is this our first "supported" .NET Engineering Services GitHub Action?  Does anybody on our team have any experience with GitHub Actions or supporting them?  Is now the time to continue to avoid them, or to being to embrace them?

missymessa: This would be our first GitHub action. I'm... on the fence about it, because the little research I did on them requires YAML on the repo to run them, which means a couple things: 

- If we have anything that's repo specific (like some of the items that Mitchell called out) that could easily be edited in the YAML. 
- Makes it easier for the YAML to get out of sync with other repos if too much is changed. 

This should definitely be a proof-of-concept for this epic, in my opinion. 

We can always fall back on webhooks (since we already have those set up) if it ends up that the GitHub action won't work for our needs.

Documentation/OnePagers/ol-maestro-deprecation.md
```diff
@@ -0,0 +1,83 @@
...
+- Should open a PR in specified 'downstream' branches when a commit to a branch is made.
+- Should should follow the similar conventions and UX as the current inter-branch merge PRs.
+- Should be reusable in any .NET repository.
+
+## Serviceability
+
+* How will the components that make up this epic be tested?
+
+The code mirror should have xunit tests written for it. These should follow a standard mockup model to avoid actual git operations. The inter-branch merge functionality may be tested in the maestro-auth-test org if necessary.
```
missymessa, line 63: Must be xunit tests? Some of our repos use nunit. Should we be more agnostic here and use whatever the repo already uses?

Documentation/OnePagers/ol-maestro-deprecation.md
missymessa: Also, with the security push going on right now, are GitHub actions still valid for us to use in this use case?

mmitche: It might be the first one we "own", but it wouldn't be the first one that the team uses. The backport PR automation, which is a GItHub Action, predates this and is in heavy use. It's also somehwat similar in functionality to inter-branch marging.

I'm not worried about GHA in this context since it's only opening PRs from A->B in github.

mmitche:
> Did you write this, or chatGPT? :D

Both.

Documentation/OnePagers/ol-maestro-deprecation.md
danmoseley: FWIW dependabot can update GH actions. I switched it on in runtime. Just noting in case that's useful.
https://github.com/dotnet/runtime/blob/main/.github/dependabot.yml#L3-L12

mmitche:
Any additional comments here?

Documentation/OnePagers/ol-maestro-deprecation.md
```diff
@@ -0,0 +1,83 @@
...
+- With one pipeline invocation per push, VM usage is higher than is probably required.
+
+A new mirror should be developed with the following qualities:
+- Should be built and run (vs. built and deployed)
+- Should be a standard .NET application that is launched by a pipeline.
+- Should deal with *all* branches and repos at the same time.
+- Should use a poll/pull model, rather than waiting for events from GitHub. This reduces onboarding cost (no need to create webhooks)
+- Should read the list of branches and repos to mirror from a repo in AzDO.
+- Should use GH and AzDO APIs to determine the state of repos on either side of the mirror to determine whether repo changes need to happen.
+- Should use a MI to do mirror pushes to AzDO and fetches from AZDO.
```
pavelsavara, line 46: What does the MI stand for here ?

missymessa: Managed Identity

PR #2971, opened 2024-05-30
Add daily secret-manager run

garath:
Related to [dotnet/dnceng#2963](https://github.com/dotnet/dnceng/issues/2963)

The reduction in maximum PAT duration requires more frequent secret-manager checkups. This PR converts the existing weekly build, which only ran secret-manager, into a daily build.

### DIFF
--- a/azure-pipelines-daily.yml
+++ b/azure-pipelines-daily.yml
@@ -1,8 +1,8 @@
 trigger: none
 
 schedules:
-- cron: 0 12 * * 1
-  displayName: Weekly Monday build
+- cron: 0 12 * * *
+  displayName: Nightly build
   branches:
     include:
     - main


dougbu (approved):
nit: might want to fix up the pipeline to reference the new file in this branch and doing a test build before merging but that's optional since you're only changing the schedule. o/c you'll break the pipeline as soon as you merge if you don't fix it up first :grin:

can this repo move to .NET 8.0 now❔

riarenas (approved):

garath:
> can this repo move to .NET 8.0 now
No, there are other constraints. It could adopt the net8 SDK but it still needs to target (or dual-target) the net6 TFM.

garath:
> it: might want to fix up the pipeline to reference the new file in this branch and doing a test build before merging but that's optional since you're only changing the schedule. o/c you'll break the pipeline as soon as you merge if you don't fix it up first

Yup, done here: [dotnet-dnceng-daily](https://dev.azure.com/dnceng/internal/_build?definitionId=1257&_a=summary)

Issue #3042, opened 2024-06-10
Bump Octokit to 12.0.0


missymessa:
Update Octokit to fix the overflow errors

### Release Note Category
- [ ] Feature changes/additions 
- [x] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
No release notes.

missymessa:
Helix Service PR: https://dnceng.visualstudio.com/internal/_git/dotnet-helix-service/pullrequest/40277

missymessa:
Both PRs merged. Closing.

PR #3043, opened 2024-06-10
Bump Octokit to 12.0.0

missymessa:
Fixes https://github.com/dotnet/dnceng/issues/3042


garath:
This might cause breaks downstream, but it's a necessary update so we'll just have to sort those as they happen.

garath (approved):

PR #3075, opened 2024-06-13
Allow multiple notification targets

dkurepa:
https://github.com/dotnet/arcade-services/issues/3633

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
@@ -173,7 +173,7 @@ public class AlertHookController : ControllerBase
 
 [Go to rule]({notification.RuleUrl})
 
-@{options.NotificationTarget}, please investigate
+{string.Join(',', options.NotificationTargets.Select(target => $"@{target}"))}, please investigate
 
 {options.SupplementalBodyText}
 
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
@@ -173,7 +173,7 @@ public class AlertHookController : ControllerBase
 
 [Go to rule]({notification.RuleUrl})
 
-{string.Join(',', options.NotificationTargets.Select(target => $"@{target}"))}, please investigate
+{string.Join(", ", options.NotificationTargets.Select(target => $"@{target}"))}, please investigate
 
 {options.SupplementalBodyText}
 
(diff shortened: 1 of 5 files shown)


src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
```diff
@@ -173,7 +173,7 @@ private NewIssue GenerateNewIssue(GrafanaNotification notification)
 
 [Go to rule]({notification.RuleUrl})
 
-@{options.NotificationTarget}, please investigate
+{string.Join(',', options.NotificationTargets.Select(target => $"@{target}"))}, please investigate
```
oleksandr-didyk, line 176: Maybe adding a space to the delimiter would make it a bit more readable? Like - `, `

oleksandr-didyk (approved):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
dkurepa: fair point, thanks

oleksandr-didyk (approved):

pavel-purma (approved):

premun (approved):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
```diff
@@ -173,7 +173,7 @@ private NewIssue GenerateNewIssue(GrafanaNotification notification)
 
 [Go to rule]({notification.RuleUrl})
 
-@{options.NotificationTarget}, please investigate
+{string.Join(", ", options.NotificationTargets.Select(target => $"@{target}"))}, please investigate
```
missymessa, line 176: So every Grafana alert is going to ping dnceng and prodcon with this change?

garath (approved):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
premun: Our team members were recently removed from dnceng so we want keep being notified

missymessa: That's going to be a lot of noise for your team when y'all don't own most of the services that are getting pings.

premun: We already get quite a lot of noise (from helix-machines or helix-service for example) but this is part of FR so we should be on these.

AlitzelMendez (approved):

Issue #3118, opened 2024-06-19
Placeholder issue to make our PR process happy


missymessa:
### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description
No release notes

PR #3148, opened 2024-06-21
A couple more removals

missymessa:
Fake issue: https://github.com/dotnet/dnceng/issues/3118

Real issue: https://dnceng.visualstudio.com/internal/_workitems/edit/6216


.vault-config/engkeyvault.yaml
garath, line 1: No secrets means this file can be entirely deleted.

garath (approved):

Issue #3172, opened 2024-06-24
AzDO Issue #6226


garath:
[Azure Boards WorkItem](https://dev.azure.com/dnceng/7ea9116e-9fac-403d-b258-b31fcf1bb293/_workitems/edit/6226)

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements

### Release Note Description
Replace some client secrets with managed identities in Grafana instances

PR #3199, opened 2024-06-28
Refactor authentication for grafana deployment tool

garath:
Resolves [dotnet/dnceng#3172](https://github.com/dotnet/dnceng/issues/3172)

Refactor authorization used in Grafana dashboard deployment to eliminate the use of service principal  passwords. Instead, the tool accepts arguments to use a Service Connection. For local dev use, it falls-back to a DefaultAzureCredential. 

The tool itself still supports Client Secrets to allow users time to migrate without build breaks.

A successful test build may be found at [Build 2484815](https://dev.azure.com/dnceng/internal/_build/results?buildId=2484815&view=logs&j=c2e05747-45ee-513b-76dd-1a76a5883b6f&t=3c6b8306-7696-5a6e-47e9-4a7eb2205652) (but be aware that the YAML is very modified to allow for testing within a PR as this path is only used within deployments).


src/Monitoring/Sdk/MonitoringPublish.cs
```diff
@@ -54,17 +58,53 @@ public sealed override bool Execute()
 
     private async Task<bool> ExecuteAsync()
     {
+        string msftTenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47";
+        TokenCredential tokenCredential;
+
+        if (ClientId == null && ServiceConnectionId == null && SystemAccessToken == null && KeyVaultServicePrincipalId == null && KeyVaultServicePrincipalSecret == null)
```
dougbu, line 64: nit: I'm not a fan of the complications in this block b/c it looks a lot like the fallbacks `DefaultAzureCredential` does automatically. we may unfortunately have no choice if current versions of `DefaultAzureCredential` don't include `AzurePipelinesCredential` yet 😦

dougbu (approved):
looks consistent to me but you may want another review. I haven't done much w/ most of the Azure.Identity classes yet

src/Monitoring/Sdk/MonitoringPublish.cs
garath: I do not think `AzurePipelineCredential` will be added to `DefaultAzureCredential`. It appears to only support `TokenCredential` classes that can quietly figure out their own configuration (which makes sense, I suppose). I looked at just building a `ChainedClientCredential` and let it figure out what was available, but the constructors get very angry if they see `null`.

PR #3248, opened 2024-07-02
Updating octokit to use latest version 13

davfost:
Updating octokit to use latest version 13

This is to try to address JSON conversion failure we are seeing in the production environment.

https://github.com/dotnet/dnceng/issues/3249

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -82,7 +82,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageVersion Include="NUnit" Version="3.13.3" />
     <PackageVersion Include="NUnit3TestAdapter" Version="3.16.1" />
-    <PackageVersion Include="Octokit" Version="12.0.0" />
+    <PackageVersion Include="Octokit" Version="13.0.0" />
     <PackageVersion Include="ServiceFabricMocks" Version="$(ServiceFabricMocksVersion)" />
     <PackageVersion Include="System.CommandLine" Version="2.0.0-beta1.21308.1" />
     <PackageVersion Include="System.Data.SqlClient" Version="4.8.6" />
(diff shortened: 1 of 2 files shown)


davfost:
Ops Issue : https://github.com/dotnet/dnceng/issues/3249

garath:
Looks like there is a breaking change in the API that needs to be resolved.

Directory.Packages.props
```diff
@@ -82,7 +82,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.2" />
     <PackageVersion Include="NUnit" Version="3.13.3" />
     <PackageVersion Include="NUnit3TestAdapter" Version="3.16.1" />
-    <PackageVersion Include="Octokit" Version="12.0.0" />
+    <PackageVersion Include="Octokit" Version="13.0.0" />
```
garath, line 85:
```suggestion
    <PackageVersion Include="Octokit" Version="13.0.1" />
```
Just noticed there was a .1 release. We may as well adopt the absolute latest while we're here.

davfost: Are you sure the new release is 'live'. It does not show up when I try to update the package in my local branch.

davfost: When I update the props file manually the local build fails because the octokit package cannot be found.

davfost: I am not updating the verions as I can't test this update locally.

garath: It is indeed published and publicly available: https://www.nuget.org/packages/Octokit/13.0.1

The restore is failing because dnceng repositories expect these sorts of packages from the from the Azure Packages mirror "[dotnet-public](https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-public)". Only Microsoft-created packages are automatically copied from NuGet to dotnet-public. Other packages must be explicitly requested via [this process](https://github.com/dotnet/arcade/blob/4a7d983f833d6b86365ea1b2b4d6ee72fbdbf944/Documentation/MirroringPackages.md). 

I suspect the the local build will succeed once that version is mirrored to dotnet-public.

garath (approved):
I'd like to see the latest Octokit get used but I won't block the PR on that as data indicates this will go a long way to solving current build problems.

Issue #3618, opened 2024-07-25
Create UploadDocs Console App for Chatbot


kcesaire:
Create a console app to automate uploading documentation to Azure blob storage

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
No release notes necessary

PR #3682, opened 2024-07-31
Created UploadDocs console app and UploadDocs.Tests

kcesaire:
Created the console app to upload markdown files from GitHub to Azure Blob Storage and created tests to test the console app. Also called the console app from the dnceng-daily-pipeline.
link to issues: [https://github.com/dotnet/dnceng/issues/3618](https://github.com/dotnet/dnceng/issues/3618) [https://github.com/dotnet/dnceng/issues/3826](https://github.com/dotnet/dnceng/issues/3826)


src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,145 @@
...
+        {
+            // Handle or log any exception that occurs during execution
+            Console.WriteLine($"An error occurred: {ex.Message}");
+        }
+    }
+
+    // Method to upload Markdown files from a GitHub repository to Azure Blob Storage
+    public async Task UploadMDFilesToBlob(string githubToken, string connectionString, string owner, string repo, string containerName)
+    {
+        try
```
AlitzelMendez, line 64: do we need the code on this try?
I think all the files are handled later on the code or are theses different files? if not I think we can delete this section

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,145 @@
...
+
+            // Initialize Azure Blob Storage client
+            BlobServiceClient blobServiceClient = new BlobServiceClient(connectionString);
+            BlobContainerClient containerClient = blobServiceClient.GetBlobContainerClient(containerName);
+
+            // Initialize a counter to track the number of uploaded files
+            int uploadCount = 0;
+
+            // Define the maximum number of pages for pagination (GitHub API limit)
+            int maxPages = 10;
```
AlitzelMendez, line 98: I think the limit is 1000 records, no 10 pages, I get that in this model 10*100 = 1000, but this makes believe that the limit is in the pages and no the records, so someone can increate the number of records per page and this is going to break, you could change your if to have something like

page*numberOfRecordsPerPage <= GitbHubLimitMaxRecords

and numberOfRecordsPerPage = 100; 
and GitbHubLimitMaxRecords should be a constant 

this could also lead to an edge case scenario in where people put a random number that never reach exactly 1000 and there are files that are not processed, I think is not that crucial because if there are more than 1000 some are not going to be processed, maybe you could order them so we always try to process at least the last ones that were updated?

if you want to avoid every scenario maybe a do while where you calculate the remaining records to pull should be better and I think would be the ideal, something like

```
int results = 0;
int recordsPerPage = 100;
int page = 1;
do {
int recordsPerPageInRun = results + recordsPerPage <= GitbHubLimitMaxRecords ? results + recordsPerPage : GitbHubLimitMaxRecords - results;

  var request = new SearchCodeRequest("extension:md repo:" + owner + "/" + repo)
                {
                    PerPage = recordsPerPageInRun,
                    Page = page
                };
....

page ++;


} while (results < GitbHubLimitMaxRecords && results.Items.Any())
```

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,145 @@
...
+using Azure.Storage.Blobs;
+using Azure.Storage.Blobs.Models;
+using Octokit;
+using Azure.Identity;
+using Azure.Security.KeyVault.Secrets;
+
+public class Program
+{
+    // Method to retrieve a list of BlobItems from a given BlobContainerClient
+    public async Task<List<BlobItem>> GetBlobItemsAsync(BlobContainerClient containerClient)
```
AlitzelMendez, line 15: are we planning to use this code?

src/Chatbot/UploadDocs/UploadDocs.csproj
```diff
@@ -2,9 +2,18 @@
...
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+
+	<ItemGroup>
```
AlitzelMendez, line 11: nit: fix indentation

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,145 @@
...
+                var result = await client.Search.SearchCode(request);
+
+                if (result.Items == null || !result.Items.Any())
+                {
+                    // Exit loop if no more items are returned
+                    break;
+                }
+
+                // Upload each Markdown file to Azure Blob Storage
+                foreach (var file in result.Items)
```
garath, line 118: What will happen if a file is deleted in the repository? It does not appear that it will be deleted in the storage container.

src/Chatbot/UploadDocs/Program.cs
kcesaire: I'm using this code for the feature I'm adding today/tomorrow to delete items in the blob that have been deleted in the GitHub repo

UploadDocs.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,167 @@
+using NUnit.Framework;
```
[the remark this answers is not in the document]
kcesaire: I will keep this one and delete the one in the folder src/Chatbot/UploadDocs.Tests/

src/Chatbot/UploadDocs/Program.cs
kcesaire: before, it didn't do anything if a file was deleted in the repository... but now I am adding code that will compare the files in the storage container to the files in the repo, and if the file isn't found in the repo it will be deleted from the storage container.

UploadDocs.Tests/UnitTest1.cs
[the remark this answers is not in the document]
kcesaire: never mind, I was talking to Cali and she reminded me that I had previously made the UploadDocs.Test folder in the wrong location of the dnceng.sln so that's why there's a duplicate I think

src/Chatbot/UploadDocs/Program.cs
kcesaire: I don't think we need it

src/Chatbot/UploadDocs/Program.cs
kcesaire: I put something similar to this in my code today.

src/Chatbot/UploadDocs/Program.cs
kcesaire: The deletion feature brought up lots of errors and issues for the code so I'm not going to implement it for the first iteration of the chatbot but for the second iteration (after the presentation) I will implement the feature.

.gitignore
```diff
@@ -361,3 +361,5 @@ MigrationBackup/
 
 # some ide stuff
 .idea
+/src/Chatbot/UploadDocs/Properties/ServiceDependencies/UploadDocsConsoleApp - Web Deploy/profile.arm.json
```
AlitzelMendez, line 364: why this needs to be ignored? same for the following line, can you explain a little bit more why we are doing this gitignore change :)

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,139 @@
...
+            String keyVaultName = "DncengChatbotKV";
+            var vaultUrl = $"https://{keyVaultName}.vault.azure.net";
+            var client = new SecretClient(vaultUri: new Uri(vaultUrl), credential: new DefaultAzureCredential());
+
+            string githubToken = (await client.GetSecretAsync("githubToken")).Value.Value;
+            string connectionString = (await client.GetSecretAsync("connectionString")).Value.Value;
+            string containerName = "automationdemo";
+            string accountName = "explorerstestdata";
+
+            // Initialize program instance
```
AlitzelMendez, line 38: can we change all this comments to Console.Writeline? 
to make this easier to debug 😄

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,139 @@
...
+            var containerEndpoint = $"https://{accountName}.blob.core.windows.net/{containerName}";
+            var containerClient = new BlobContainerClient(new Uri(containerEndpoint), new DefaultAzureCredential());
+
+            // Retrieve existing blobs
+            var existingBlobs = await GetBlobItemsAsync(containerClient);
+            var existingBlobPaths = new HashSet<string>(existingBlobs.Select(b => b.Name));
+
+            // Counters for tracking
+            int uploadedFilesCount = 0;
+            // int deletedFilesCount = 0; // Counter for deleted files removed
```
AlitzelMendez, line 74: nit: this is not used, lets remove the line

.gitignore
kcesaire: These are the paths to the files made when I was trying to publish the console app as a web job (before you told me I didn't need to), I clicked ignore files on Visual Studio instead of deleting them because I didn't know there was a difference. Should I delete the two files and then these two lines?

AlitzelMendez: yes, delete the files and the two lines please 😄

missymessa (approved):
LGTM shipit

src/Chatbot/UploadDocs/Program.cs
```diff
@@ -1,2 +1,154 @@
...
+                    uploadedFilesCount++;
+                    Console.WriteLine($"Uploaded {repo}/{file.Path}");
+                }
+
+                totalResults += result.Items.Count;
+                page++;
+
+            } while (totalResults < GitHubLimitMaxRecords && result.Items.Any());
+
+            // NOTE: Code for deleting files from blob storage that were deleted in GitHub has been removed
```
garath, line 136: Just to be clear, this means that files in blob storage never gets deleted, so the bot will always include that data in its context?

PR #3699, opened 2024-08-01
Rollout scorer using MI to access azure storage table

AlitzelMendez:
We already have a MI installed on the azure function that is used around the code, so I took advantage of that. 

Created ScorecardEntity : ITableEntity because I need to update to ITableEntity, the model lives in other repo: https://github.com/dotnet/dnceng-shared/blob/77fb751f40c92262450af10314e965a1a708857b/src/Microsoft.DotNet.Services.Utility/AzureStorageTableEntities.cs#L44

but is not used in any other place, so it doesn't need to be in other place, I will triple check this and remove the model for the other repository

### DIFF
--- a/src/RolloutScorer/RolloutScorer/Models/ScorecardEntity.cs
+++ b/src/RolloutScorer/RolloutScorer/Models/ScorecardEntity.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using Azure;
+using ITableEntity = Azure.Data.Tables.ITableEntity;
+
+namespace RolloutScorer.Models;
+
+public class ScorecardEntity : ITableEntity
+{
+    public ScorecardEntity()
+    {
+    }
+
+    public ScorecardEntity(DateTimeOffset date, string repo)
+    {
+        PartitionKey = date.ToString("yyyy-MM-dd");
+        RowKey = repo;
+    }
+    
+    public string PartitionKey { get; set; }
+    public string RowKey { get; set; }
+    public DateTimeOffset? Timestamp { get; set; }
+    public ETag ETag { get; set; }
+
+    public int TotalScore { get; set; }
+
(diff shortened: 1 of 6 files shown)


src/RolloutScorer/RolloutScorer/Models/ScorecardEntity.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using Azure;
+using ITableEntity = Azure.Data.Tables.ITableEntity;
+
+namespace RolloutScorer.Models;
+
+public class ScorecardEntity : ITableEntity
```
mmitche, line 7: It doesn't look like this is referenced in dnceng-shared. Remove it from there?

src/RolloutScorer/RolloutScorer/Utilities.cs
```diff
@@ -72,13 +70,16 @@ public static string HandleApiRedirect(HttpResponseMessage redirect, Uri apiRequ
...
     }
+    public static TableClient GetScorecardsCloudTable()
+    {
+        TableClient tableClient = new TableClient(GetStorageTableEndPoint(), ScorecardsStorageAccount.ScorecardsTableName, new DefaultAzureCredential());
+
+        return tableClient;
+    }
 
-    public static CloudTable GetScorecardsCloudTable(string storageAccountKey)
+    private static Uri GetStorageTableEndPoint()
```
mmitche, line 80: Theres only one reference to GetStorageTableEndpoint(). suggest just inlining this into the caller.

mmitche (approved):
Approved with minor comments.

mmitche:
There's also `RolloutScorerFunction`, which has some SAS uri usage

garath (approved):

mmitche (approved):

garath (approved):

PR #3763, opened 2024-08-06
Fix auth errors while running locally

mmitche:
Fixes: https://github.com/dotnet/dnceng/issues/3764

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Program.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Program.cs
@@ -20,7 +20,7 @@ public static class Program
             .ConfigureWebHostDefaults(host =>
             {
                 host.UseStartup<Startup>()
-                    .UseUrls("http://localhost:5000/")
+                    .UseUrls("https://localhost:5000/")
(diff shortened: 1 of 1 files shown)


mmitche:
Much thanks to @garath for figuring out the root issue here.

AlitzelMendez (approved):

src/DotNet.Status.Web/DotNet.Status.Web/Program.cs
```diff
@@ -20,7 +20,7 @@ public static void Main(string[] args)
             .ConfigureWebHostDefaults(host =>
             {
                 host.UseStartup<Startup>()
-                    .UseUrls("http://localhost:5000/")
+                    .UseUrls("https://localhost:5000/")
```
garath, line 23: Does this get overridden to http when deployed to the app service? The site sits behind an Application Gateway that handles the TLS termination so non-http is required by the service itself.

mmitche: It doesn't look to be overridden by any deployment setting. So my guess is no. Good call. I'll use your fix instead.

PR #3774, opened 2024-08-07
Move DotNet.Status.Web to MI

mmitche:
Resolves: https://github.com/dotnet/dnceng/issues/3776

- Update to Azure.Data.Tables
- Use managed identities in place of SAS tokens.

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/AzureTableTokenStore.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/AzureTableTokenStore.cs
@@ -49,7 +49,7 @@ public class AzureTableTokenStore : ITokenStore, ITokenRevocationProvider
         }
         else
         {
-            table = new TableClient(new Uri(options.TableUri, UriKind.Absolute), options.TableName, new DefaultAzureCredential());
+            table = new TableClient(new Uri(options.TableUri, UriKind.Absolute), options.TableName, new ManagedIdentityCredential());
         }
         return table;
     }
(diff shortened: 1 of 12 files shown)


mmitche:
Still doing a few more bits of verification on the deployment table side of this, and I want to update the use of DefaultAzureCredential() to use the system assigned identities in non-dev mode.

mmitche:
Ready for review @garath @AlitzelMendez. I've verified that perms are set appropriately for the system managed identity in the staging resources.

src/DotNet.Status.Web/DotNet.Status.Web/AzureTableTokenStore.cs
```diff
@@ -137,37 +136,24 @@ long MakeTokenId()
...
-
-        List<StoredTokenData> tokens = new List<StoredTokenData>();
-        do
-        {
-            TableQuerySegment<TokenEntity> segment = await table.ExecuteQuerySegmentedAsync(query, continuationToken, requestOptions: null, operationContext: null, cancellationToken);
-            continuationToken = segment.ContinuationToken;
-            tokens.AddRange(segment.Results.Select(Return));
-        } while (continuationToken != null);
+        TableClient table = await GetCloudTable();
+        var results = table.QueryAsync<TokenEntity>(entity => entity.PartitionKey == userId.ToString());
```
AlitzelMendez, line 154: are we iterating through all the pages? my understanding is that in this case it will only return one page with 1000 records which I think it is more than enough for this query, but something to keep in mind

AlitzelMendez (approved):

src/DotNet.Status.Web/DotNet.Status.Web/AzureTableTokenStore.cs
mmitche: I think based on the API docs: https://learn.microsoft.com/en-us/dotnet/api/azure.asyncpageable-1.getasyncenumerator?view=azure-dotnet, we should be.

`AsyncPageable`'s GetAsyncEnumerator indicates that it may take multiple service requests to resolve. I think that would end up getting called in this case.

Good call though. I'm going to give this a shot and see what happens.

mmitche: Confirmed that this does end up evaluating all pages.

PR #3794, opened 2024-08-09
Support PAT-backed AzDO Service Connections

garath:
Resolves [dotnet/dnceng#2361](https://github.com/dotnet/dnceng/issues/2361)

Support Azure Devops service connections in secret-manager by

- Add "service connections" as a storage destination for secrets
- Add a "service connection" type that represents individual connections and their configuration


src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionVault.cs
```diff
@@ -0,0 +1,123 @@
...
+public class ServiceConnectionVault : StorageLocationType<ServiceConnectionVaultParameters>
+{
+    private readonly ILogger<ServiceConnectionVault> _logger;
+    private readonly IServiceProvider _serviceProvider;
+
+    public ServiceConnectionVault(ILogger<ServiceConnectionVault> logger, IServiceProvider serviceProvider)
+    {
+        _logger = logger;
+        _serviceProvider = serviceProvider;
+
```
missymessa, line 40: nit: so much space for activities

missymessa (approved):
LGTM shipit

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionVault.cs
garath: Fixed in 866365f

.vault-config/service-connections-dnceng.yaml
```diff
@@ -0,0 +1,117 @@
...
+    type: azure-devops-service-endpoint
+    parameters:
+      authorization:
+        type: azure-devops-access-token
+        parameters:
+          domainAccountName: dn-bot
+          domainAccountSecret:
+            location: helixkv
+            name: dn-bot-account-redmond
+          organizations: dnceng devdiv microsoft dotnet-security-partners
```
riarenas, line 116: I don't think the list of organizations here is correct. This is a service connection in dnceng, connected to a single feed, so it should not need all org access

.vault-config/service-connections-dnceng.yaml
```diff
@@ -0,0 +1,117 @@
...
```
riarenas, line 103: This PAT only needs the last org in this list. This is a nuget service connection connected to a single feed in that org.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -20,5 +26,28 @@ protected override void ConfigureServices(IServiceCollection services)
...
+                else if (handlerBuilder.PrimaryHandler is HttpClientHandler httpClientHandler)
+                {
+                    httpClientHandler.CheckCertificateRevocationList = true;
+                }
+                else
+                {
+                    throw new InvalidOperationException($"Could not create client with CRL check, HttpMessageHandler type {handlerBuilder.PrimaryHandler.GetType().FullName ?? handlerBuilder.PrimaryHandler.GetType().Name} is unknown.");
+                }
+            });
+        });
```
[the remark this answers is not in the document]
riarenas: Ah, we do this as well in helix-service. Carry on.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceConnectionSecretType.cs
```diff
@@ -1,64 +0,0 @@
-﻿using System;
```
riarenas, line 1: I couldn't find any usage of this secret type, but we should keep an eye for failures related to this deletion when we adopt this version of secret-manager.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceEndpointSecretType.cs
```diff
@@ -0,0 +1,47 @@
...
+        public AuthorizationParameters Authorization { get; set; }
+
+        public class AuthorizationParameters
+        {
+            public string Type { get; set; }
+            public AzureDevOpsAccessToken.Parameters Parameters { get; set; }
+        }
+    }
+
+    // This secret type uses other secret types to perform actions. It gets access to those types via SecretTypeRegistry. Due to complications in how SecretTypeRegistry is constructed, we must use IServiceProvider get an instance of SecretTypeRegistry at runtime rather than inject directly to avoid creating a circular dependency.
```
riarenas, line 30: In your opinion, Is this something we should look into improving as a follow up issue?

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpoint.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        Id = id;
+        Name = name;
+        Description = description;
+        Type = type;
+    }
+
+    internal static ServiceEndpoint Create(AdoServiceEndpoint adoServiceEndpoint)
+    {
+        // The JSON schema allows more flexibility than used for the application model. We assume that there is only one element in the serviceEndpointProjectReferences array. We assume that the "name" and "description" elements in that array are what drive the info shown in the AzDO UI and are
```
riarenas, line 24: Don't leave me hanging!

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,202 @@
...
+        _tokenCredentialProvider = credentialProvider;
+    }
+
+    private async Task GetOrUpdateTokenAsync(CancellationToken cancellationToken = default)
+    {
+        if (_currentAccessToken is null || _currentAccessToken.Value.RefreshOn < DateTimeOffset.UtcNow)
+        {
+            TokenCredential tokenCredential = await _tokenCredentialProvider.GetCredentialAsync();
+            // 499b84ac-1321-427f-aa17-267ca6975798 is the Azure DevOps API application
+            // https://ms.portal.azure.com/#view/Microsoft_AAD_IAM/ManagedAppMenuBlade/~/Overview/objectId/71dba5a0-a77c-4b64-bcc4-f5f98be267fe/appId/499b84ac-1321-427f-aa17-267ca6975798
```
riarenas, line 61: [nit] wondering if linking to a public document would be better than the azure object? Something like https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/?view=azure-devops-rest-7.1&tabs=powershell includes this GUID.

riarenas: Also, thoughts on extracting as a constant?

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,202 @@
...
+            _httpClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Basic",
+            Convert.ToBase64String(
+                Encoding.ASCII.GetBytes($":{_currentAccessToken.Value.Token}")));
+        }
+    }
+
+    // https://learn.microsoft.com/en-us/rest/api/azure/devops/serviceendpoint/endpoints/get?view=azure-devops-rest-7.2&tabs=HTTP
+    public async Task<ServiceEndpoint> Get(Guid id, CancellationToken cancellationToken = default)
+    {
+        Console.WriteLine("Whoa!");
```
riarenas, line 74: Not opposed, but debug statement?

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,202 @@
...
+                }
+            }
+            else
+            {
+                throw new ServiceEndpointClientException("'serviceEndpointProjectReferences' is not a valid JsonArray.");
+            }
+
+        }
+
+        // TODO: Should I validate the JsonNode schema? If so, do it everywhere.
```
riarenas, line 160: Did you decide? followup needed?

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+            else
+            {
+                throw new ServiceEndpointClientException("Neither 'nugetkey' nor 'apikey' found in the authorization parameters.");
+            }
+        }
+
+        Uri updateEndpoint = CreateUpdateUri(config, id, _apiVersion);
+
+        //using HttpResponseMessage postResponseBody = await _httpClient.PutAsJsonAsync(updateEndpoint, getResponseBody, cancellationToken);
```
riarenas, line 179: Delete

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionMagicString.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DncEng.SecretManager.StorageTypes;
+
+public class ServiceConnectionMagicString
```
riarenas, line 6: [nit] This looks like a helper to strictly build service connection descriptions. `MagicString` made it sound like something more generic.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionVault.cs
```diff
@@ -0,0 +1,121 @@
...
+[Name("azure-devops-project")]
+/// <summary>
+/// This represents a collection of Azure DevOps Service Connections, also called Service Endpoints, that are managed by secret-manager. 
+/// 
+/// Service Connection are grouped in an Azure DevOps project. Not all service connections in a project need be managed by secret-manager; unrecognized service connections are ignored. 
+/// 
+/// Service Connections are recognized by secret-manager by well-known text in their description.
+/// </summary>
+public class ServiceConnectionVault : StorageLocationType<ServiceConnectionVaultParameters>
```
riarenas, line 31: What is the meaning of a vault in this context?

riarenas (approved):
Just some minor questions!

.vault-config/service-connections-dnceng.yaml
garath: Possible, sure. The data on these is a little... loose. I'll update.

.vault-config/service-connections-devdiv.yaml
garath, line 1: Does anyone have any opinion about these manifest files being in this repository? The only other place that makes sense to me is Arcade (since they're almost all related to customer builds.)

riarenas: I think this is a good place to centrally manage them.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceEndpointSecretType.cs
garath: No, not really. It's just not a serious enough caveat in my opinion.

riarenas:
Oh, do we want to add a scenario test for this? (I think we do!) can you open a follow-up issue?

.vault-config/service-connections-dnceng.yaml
garath: Fixed in 105e07c

.vault-config/service-connections-dnceng.yaml
garath: Fixed in 105e07c

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/ServiceConnectionSecretType.cs
garath: This was an old attempt to make service-connections easier to manage I made a while back. It never caught on. This new system fully replaces it.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionVault.cs
garath: To me, it's just a word to describe the place secrets are being written to. I took it as a pattern for secret-manager: it stores secrets in "vaults". Happy to entertain other names, this is just what came to me.

riarenas: So a service connection is a vault for PATs? I can see it that way.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: URL fixed in 642c08c2.

I opted to not extract a constant as there didn't seem to be a good upper-level to put it, and didn't really remove all that many instances of its use. So, it was a grey area for me that I decided to leave alone.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: Fixed in d2b56d62. I probably shouldn't post PRs when I'm tired.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: Fixed in 5e744b62. Decided to leave as-is. My best understanding now says this will be fine. If problems occur, then that will be good info for informing a round of robustness improvements.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: Fixed in 5e744b6

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionMagicString.cs
garath: As it is, it is just for service connections. I could not come up with a reasonable plan to make it properly generic. I view the "StorageConnection" prefix in the name to be enough to say its specialized. Happy to entertain other name suggestions.

riarenas: I was thinking making it more related to the description. it's basically a description builder + parser right?

garath: Oh, sure. Like `ServiceConnectionDescriptionBuilder`? Doesn't quite cover the 'parser' part but it seems maybe net-better?

riarenas: Yep I'm good with that! that was one of my options, aside from the dreaded "Manager" suffix.

riarenas (approved):

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/SynchronizeCommand.cs
```diff
@@ -55,6 +55,7 @@ public override async Task RunAsync(CancellationToken cancellationToken)
     {
         try
         {
+            Console.OutputEncoding = System.Text.Encoding.UTF8;
```
davfost, line 58: Can you comment why this encoding is needed, it helps ensure people don't remove it by accident at a later date.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -20,5 +26,28 @@ protected override void ConfigureServices(IServiceCollection services)
         services.AddSingleton<SettingsFileValidator>();
         services.AddNamedFromAssembly<SecretType>(Assembly.GetExecutingAssembly());
         services.AddNamedFromAssembly<StorageLocationType>(Assembly.GetExecutingAssembly());
+
+        services.AddHttpClient();
+        services.Configure<HttpClientFactoryOptions>(factory =>
```
davfost, line 31: Can you add a comment as to why HttpClient needs handler overrides for this process

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/AzureDevOpsAccessToken.cs
```diff
@@ -40,45 +39,21 @@ public AzureDevOpsAccessToken(ISystemClock clock, IConsole console)
...
-        {
-            var errorContent = await response.Content.ReadAsStringAsync(cancellationToken);
-            if (errorContent.Contains("AADSTS50079"))
-            {
-                throw new Exception("Failed to get a token from AzDO. Please connect to CORPNET and try again." + Environment.NewLine + Environment.NewLine + errorContent);
-            }
-
-            throw new Exception(errorContent);
-        }
+        UsernamePasswordCredential credential = new(userName, password, msftTenantId, ClientId);
```
davfost, line 49: Is there an advantage to using this variable construction patter instead of the var variable = object pattern?

The 'var' patter is meant to reduce the amount of code change required when object types change in a project. In that patter objects are ideally generated by factories, or pass down though methods. This ensure the only place changes are needed are in the factories or in the method definitions.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    private async Task GetOrUpdateTokenAsync(CancellationToken cancellationToken = default)
+    {
+        if (_currentAccessToken is null || _currentAccessToken.Value.RefreshOn < DateTimeOffset.UtcNow)
+        {
+            TokenCredential tokenCredential = await _tokenCredentialProvider.GetCredentialAsync();
+            // 499b84ac-1321-427f-aa17-267ca6975798 is the Azure DevOps API application
+            // https://learn.microsoft.com/en-us/rest/api/azure/devops/tokens/?view=azure-devops-rest-7.1&tabs=powershell 
+            TokenRequestContext requestContext = new(["499b84ac-1321-427f-aa17-267ca6975798"]);
```
davfost, line 62: Since this is basically a magic string GUID id can it be stored in a local named constant with he comments above.
It is less likely that someone alter a cost value and forget to alter the comments if it is a class constant/private value.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionMagicString.cs
davfost: Just seconding Ricardo's comments 'magic strings' to me means string that are statically defined for an critical value and without explanation in a method body.

davfost (dismissed):

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/AzureDevOpsAccessToken.cs
garath: Personally I am quite against `var`. It hides the type and makes code difficult to follow. Target-typed new is much better.

But, really, I do this because the default linter rules prefer it. (And since I do too, I don't complain.)

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
```diff
@@ -0,0 +1,198 @@
...
+    public ServiceEndpointClient(HttpClient httpClient, IOptions<Configuration> config, ITokenCredentialProvider credentialProvider)
+    {
+        _httpClient = httpClient;
+        this.config = config.Value;
+        _tokenCredentialProvider = credentialProvider;
+    }
+
+    private async Task GetOrUpdateTokenAsync(CancellationToken cancellationToken = default)
+    {
+        if (_currentAccessToken is null || _currentAccessToken.Value.RefreshOn < DateTimeOffset.UtcNow)
```
premun, line 57: Don't all credentials already do that? Seems like `SecretManagerCredentialProvider` only delegates to the usual ones and those do it always - you just keep calling `GetTokenAsync` and they refresh when needed, as long as you keep the instance around.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/SynchronizeCommand.cs
garath: This is actually unrelated to the feature. It fixes a bug that can be seen in the text output from secret-manager. Switching the console to UTF8 is what allows the emojis to display properly which currently fails (meaning, they look bad) during the first set of data displayed.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
garath: The handler ensures that certificate revocation is checked according to Microsoft policy.

garath:
> Oh, do we want to add a scenario test for this? (I think we do!) can you open a follow-up issue?

Test added in 1d6181d and refined in subsequent commits.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/ServiceConnectionMagicString.cs
garath: Renamed in b1ed166

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: Each call to `GetTokenAsync` gets a fresh token. I was trying to reduce the number of network calls made (but admit it probably does not have a significant impact in this situation).

garath: This is how I interpreted the text in the "Remarks" section for that method: [gettokenasync#remarks](https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential.gettokenasync?view=azure-dotnet#remarks)

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpoint.cs
garath: Clarified in 88cfeb3

garath:
Note that I modified the CI YAML in c38ec17 to allow me to run the scenario-tests outside of an actual deployment. That will be reverted before this PR is merged; it's only here so I can prove it works.

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
premun: That's interesting, I see conflicting information elsewhere. I read that it's not even possible to turn the cache off unless you create a new credential object:
https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/samples/TokenCache.md#disable-caching

> In-memory token caching is the default option provided by the Azure Identity library. This caching approach allows apps to store access tokens in memory. With in-memory token caching, the library first determines if a valid access token for the requested resource is already stored in memory. If a valid token is found, it's returned to the app without the need to make another request to Microsoft Entra ID. If a valid token isn't found, the library will automatically acquire a token by sending a request to Microsoft Entra ID.

or here

https://learn.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential.gettokenasync?view=azure-dotnet#definition

> Sequentially calls [GetToken(TokenRequestContext, CancellationToken)](https://learn.microsoft.com/en-us/dotnet/api/azure.core.tokencredential.gettoken?view=azure-dotnet#azure-core-tokencredential-gettoken(azure-core-tokenrequestcontext-system-threading-cancellationtoken)) on all the included credentials in the order [EnvironmentCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.environmentcredential?view=azure-dotnet), [ManagedIdentityCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.managedidentitycredential?view=azure-dotnet), [SharedTokenCacheCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.sharedtokencachecredential?view=azure-dotnet), and [InteractiveBrowserCredential](https://learn.microsoft.com/en-us/dotnet/api/azure.identity.interactivebrowsercredential?view=azure-dotnet) returning the first successfully obtained [AccessToken](https://learn.microsoft.com/en-us/dotnet/api/azure.core.accesstoken?view=azure-dotnet). Acquired tokens are cached by the credential instance. Token lifetime and refreshing is handled automatically. Where possible, reuse credential instances to optimize cache effectiveness.

premun: Maybe the `TokenCredential` you found is an abstract parent that does not cache but the implementations do? That would make sense?

garath: Interesting. TokenCredential is the abstract class the rest of those implement, so I'm open to the idea that its documentation is just guidance for everyone else. I'll poke through the source code and see what's actually happening.

garath:
> Note that I modified the CI YAML in [c38ec17](https://github.com/dotnet/dnceng/commit/c38ec17c3fd4fb672a9999f89200b94389095fdb) to allow me to run the scenario-tests outside of an actual deployment. That will be reverted before this PR is merged; it's only here so I can prove it works.

This was reverted in c10f6e6 and is ready for another round of reviews.

riarenas:
It looks like there are test failures

garath:
> It looks like there are test failures

Sort of, but yes. The tests pass locally, and the CI output looks like the tests are as expected but AzDO is misinterpreting the text as a failure. I'm guessing this is due to me bumping the NUnit library version. Since it is only showing in AzDO I am using a different branch for experiments so I can run the pipeline.

garath:
> Sort of, but yes. The tests pass locally, and the CI output looks like the tests are as expected but AzDO is misinterpreting the text as a failure. I'm guessing this is due to me bumping the NUnit library version. Since it is only showing in AzDO I am using a different branch for experiments so I can run the pipeline.

I've narrowed this to occuring with NUnitTestAdapter 4.5.0 and 4.6.0 (the latest). It seems to be related to [changes made to the ConsoleOut default value](https://docs.nunit.org/articles/vs-test-adapter/AdapterV4-Release-Notes.html#bug-fixes) (this lists it as a change in 4.6.0 but I see the same effects in 4.5.0). 

I'm opting to fix this by forcing the old default (change in 92432869).

src/SecretManager/Microsoft.DncEng.SecretManager/ServiceConnections/ServiceEndpointClient.cs
garath: I'm going to leave this as-is. It works and seems safer.

.vault-config/service-connections-devdiv.yaml
```diff
@@ -0,0 +1,65 @@
...
+    parameters:
+      authorization:
+        type: azure-devops-access-token
+        parameters:
+          domainAccountName: dn-bot
+          domainAccountSecret:
+            location: helixkv
+            name: dn-bot-account-redmond
+          organizations: dnceng
+          scopes: packaging_write
```
riarenas, line 65: There are a few more devdiv service connections in the https://dev.azure.com/dnceng/internal/_wiki/wikis/DNCEng%20Services%20Wiki/1249/PAT-recycling list

* dnceng-dotnet8-workloads
* dnceng-dotnet9-workloads
* dotnet-tools-dnceng-public-feed
* dnceng/internal

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/ServiceConnectionTests.cs
```diff
@@ -0,0 +1,169 @@
...
+{
+    [TestFixture]
+    [Category("PostDeployment")]
+    [Explicit("requires restricted resources, see inline comments")]
+    public class ServiceConnectionTests : ScenarioTestsBase
+    {
+        /// <summary>
+        /// This performs an end-to-end of service connection support by generating a new token, applying it to the service connection, and queuing a build that uses the service connection. The result of the build determines the result of this test.
+        /// 
+        /// Note: This test is marked "Explicit" so it does not run in regular CI. It requires access to dn-bot's credentials, which, to reduce security footprint, is not available to the scenario-test identity. This may be run locally by any developer with access to the necessary vaults.
```
riarenas, line 24: IF this test is not exercised during automation regularly, how will we make sure it continues to work?

garath: Yeah this is a new pattern. I don't love that it can't be automatically run, but I also end up thinking there is enough value to keep it available. My best idea right now is that reviewers will have to help out make sure to call it out whenever it might be useful. 

Obviously, I'm open to better ideas.

riarenas: My experience is that that model doesn't quite work. I would rather we design a scenario test that will run during CI. We can check this in for now but there should be a followup.

garath: I agree with your concerns and am happy for a follow-up. 

For future cause, the technical problem that blocked a better solution is that only "User" entities can generate PATs. Service Principals and Managed Identities cannot. So, a proper test would require a test-only User (and I deemed this too much overhead for the effort at this time).

.vault-config/service-connections-devdiv.yaml
garath: Added in 6973830

riarenas (approved):

PR #3996, opened 2024-09-06
Fix dotnet version install in CI and daily pipelines

garath:
Follow-up to dotnet/dnceng#3794. Fixes [dotnet/dnceng#3989](https://github.com/dotnet/dnceng/issues/3989) and fixes dotnet/dnceng#3983.

Ensure net6 is available for secret-manager runs in CD and daily pipelines


missymessa (approved):
LGTM shipit

PR #4183, opened 2024-09-27
Add Audit Logging To SecreteManager.exe

davfost:
Add Audit Logging To SecreteManager.exe

- Adds the OpenTelemetry.Audit.Geneva package and dependencies
- Following logging recommendations from Geneva team
- Added non-volatile logging operations to secret update and create operations
- Adds new command line options for setting the target service ID to used when performing audit logging

Part Of GitHub Issue https://github.com/dotnet/arcade/issues/13217


src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -1,25 +1,53 @@
+using Microsoft.DncEng.CommandLineLib;
```
dougbu, line 1: I'm not really reviewing this, knowing very little about Geneva. I would still generally prefer we stick w/ `using System*`s first or at least leave existing files alone in `using` order.

davfost: Hi @dougbu Are there tools that ensure that ordering logic like the ones that exist to 'remove and sort' in Visual Studio?

If so, do we have any documentation on how to apply them in Visual Studio so these ordering rules can be globally applied?

dougbu: from what I've seen most of us have the "sort system directives first" setting enabled in VS or VS Code. it's not necessary to use an editor config file for that but we could add such a file if we ever made the files consistent. for now, you could just undo the "sort" part of "remove and sort"

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/InfoCommand.cs
```diff
@@ -3,21 +3,25 @@
...
 
-    public InfoCommand(IConsole console)
+    public InfoCommand(GlobalCommand globalCommand, IConsole console): base(globalCommand)
     {
         _console = console;
     }
 
     public override Task RunAsync(CancellationToken cancellationToken)
     {
+        // Provides a curtisy warning message if the ServiceTreeId option is set to a empty guid
```
mmitche, line 22:
```suggestion
        // Provides a courtesy warning message if the ServiceTreeId option is set to a empty guid
```

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// <summary>
+        /// Overides the GetOptions method from the base class to add a cusotm option for the ServiceTreeId
+        /// </summary>
+        public override OptionSet GetOptions()
+        {
+            return new OptionSet()
+            {
+                {"servicetreeid=", "The service tree ID (must be a valid GUID id from aka.ms/servicetree)", id =>
+                    {
+                        if (Guid.TryParse(id, out var guid))
```
mmitche, line 49: Since someone took the time to specify a service tree ID here, I would just error out if it's not a valid GUID.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        /// <summary>
+        /// Base constructor for the SecurityAuditLogger
+        /// </summary>
+        public SecurityAuditLogger(Guid serviceTreeId)
+        {
+            var auditFactory = AuditLoggerFactory.Create(options =>
+            {
+                // We use ETW as the destination for the audit logs becsue the application is not gurenteed to run on windows
+                options.Destination = AuditLogDestination.ETW;
```
mmitche, line 28: ETW is available on Linux?

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
```diff
@@ -1,32 +1,33 @@
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.DncEng.CommandLineLib;
+using Microsoft.VisualStudio.Services.Common;
```
mmitche, line 4: Necessary?

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/AzureKeyVault.cs
```diff
@@ -107,44 +110,106 @@ private static ImmutableDictionary<string, string> GetTags(global::Azure.Securit
 
     public override async Task SetSecretValueAsync(AzureKeyVaultParameters parameters, string name, SecretValue value)
     {
-        SecretClient client = await CreateSecretClient(parameters);
-        var createdSecret = await client.SetSecretAsync(name, value.Value ?? "");
-        var properties = createdSecret.Value.Properties;
-        foreach (var (k, v) in value.Tags)
+        // The default audit state should alwasy be failure and overwritten with success at the end of the operation.
```
garath, line 113:
```suggestion
        // The default audit state should always be failure and overwritten with success at the end of the operation.
```

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+            {
+                LogSecretAction(OperationType.Update, operationName, credentialProvider, secretName, secretStoreType, secretLocation, result, resultMessage);
+            }
+            // Audit logging is a 'volitile' operation meaning it can throw exceptions if logging fails.
+            // This could lead to service instability caused by simple logging issues which is not desirable.
+            // So we catch all exceptions and write them to console as a last resort.
+            // The hope is that app insights will also catch the base exception for debugging.
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Failed to add audit log for secret update!: <{ex.Message}>");
```
garath, line 55: Will the audit message always be safe for public view in a console?

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+using OpenTelemetry.Audit.Geneva;
+
+namespace Microsoft.DncEng.SecretManager
+{
+    /// <summary>
+    /// SecurityAuditLogger is a class that is used to log security audit events to the security event log and Geneva
+    /// </summary>
+    public class SecurityAuditLogger
+    {
+        private ILogger ControlPanelLogger;
```
garath, line 16:
```suggestion
        private ILogger ControlPlaneLogger;
```
I think that's what you meant?

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// <summary>
+        /// Add an audit log for secret update operations perfomred on behalf of a user.
+        /// </summary>
+        public void LogSecretUpdate(ITokenCredentialProvider credentialProvider, string secretName, string secretStoreType, string secretLocation, OperationResult result = OperationResult.Success, string resultMessage = "", [CallerMemberName] string operationName = "")
+        {
+            try
+            {
+                LogSecretAction(OperationType.Update, operationName, credentialProvider, secretName, secretStoreType, secretLocation, result, resultMessage);
+            }
+            // Audit logging is a 'volitile' operation meaning it can throw exceptions if logging fails.
```
garath, line 49:
```suggestion
            // Audit logging is a 'volatile' operation meaning it can throw exceptions if logging fails.
```

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+
+        internal void LogSecretAction(OperationType operationType, string operationName, ITokenCredentialProvider credentialProvider, string secretName, string secretStoreType, string secretLocation, OperationResult result, string resultMessage)
+        {
+            // Perform a no op if audit logging is suppressed
+            if (SuppressAuditLogging)
+            {
+                return;
+            }
+
+            // The token applicatoin id of the client running the assembly.
```
garath, line 67:
```suggestion
            // The token application id of the client running the assembly.
```

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+            // NOTE: The user identity here should be something 'dynamic'.
+            // If you are hard coding this value you should question if this Audit Log is useful
+            // as it is likly redundant to lower level permission change logging that is already occuring.
```
garath, line 70: Should these comments be here? They look like copy-paste from an example?

NuGet.config
```diff
@@ -1,7 +1,8 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
   <packageSources>
     <clear />
+    <add key="AzureGenevaMonitoring" value="https://msblox.pkgs.visualstudio.com/_packaging/AzureGenevaMonitoring/nuget/v3/index.json" />
```
garath, line 5: What permissions are required to access this feed? Will there be issues with CI and individual users getting access?

davfost: There is currently an issue with this feed and it blocked the PR build from completing. I have been discussing it with Matt Mitchell to see how it can be resolved. As you observed this feed requires external permissions that are not granted so the package restore is failing.

The basic issue is that the Audit team says we have to use the version of the package that comes from 'this' feed and not the public version of the package. 

- I need to investigate if the 'public' feed version of the package has the same class and methods defined. If so, we can do something where the 'internal feed' is different than the feed used in the 'public' processes. 
- I also have open questions to the Audit team to verify if the package from 'this' feed CAN be exposed publicly. If so we may be able to just manually move target versions into one of the feeds we manage directly.

Matt said these two options have been used in the past to deal with other packages that have similar issues.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
garath: +1. We also don't typically capture ETW data. Where does audit information ultimately end up?

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+        private ILogger ControlPanelLogger;
+
+        private bool SuppressAuditLogging = false;
+
+        /// <summary>
+        /// Base constructor for the SecurityAuditLogger
+        /// </summary>
+        public SecurityAuditLogger(Guid serviceTreeId)
+        {
+            var auditFactory = AuditLoggerFactory.Create(options =>
```
garath, line 25: What does this ILogger do if the Service ID is `Guid.Empty`? Does it still try to log or is that enough to make it silent?

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
```diff
@@ -26,6 +24,11 @@ private async Task TestCommand(DateTimeOffset now, string manifestText, string l
         var cancellationToken = cts.Token;
 
         var services = new ServiceCollection();
+        // Dependency injection instruction needed to support properties used for Geneval Logging operations
+        services.AddSingleton(new GlobalCommand());
+        services.AddSingleton(new SecurityAuditLogger(Guid.Empty));
```
garath, line 29: Since these are test classes, would it be better to make a do-nothing version of SecurityAuditLogger? (I'm not sure what SecurityAuditLogger does.)

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
```diff
@@ -26,6 +24,11 @@ private async Task TestCommand(DateTimeOffset now, string manifestText, string l
...
+
+        // Original dependency injection instructions
```
garath, line 31: I don't thin kthis comment (or the previous comment) is necessary. The code shouldn't have a sense of "human time". That's what version control handles. The code should always just be "this is what it is right now".

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateAllCommand.cs
```diff
@@ -4,8 +4,10 @@
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+
 using Microsoft.DncEng.CommandLineLib;
 using Microsoft.DncEng.SecretManager.StorageTypes;
+using Microsoft.VisualStudio.Services.Common;
```
garath, line 10: What's bringing this in?

src/SecretManager/Microsoft.DncEng.SecretManager/ITokenCredentialProvider.cs
```diff
@@ -5,5 +5,15 @@ namespace Microsoft.DncEng.SecretManager;
 
 public interface ITokenCredentialProvider
 {
+    /// <summary>
+    /// The applicatoin ID for the credential provider.
```
garath, line 9:
```suggestion
    /// The application ID for the credential provider.
```

Directory.Packages.props
```diff
@@ -57,15 +57,15 @@
     <PackageVersion Include="Microsoft.DotNet.SwaggerGenerator.MSBuild" Version="$(MicrosoftDotNetSwaggerGeneratorMSBuildVersion)" />
     <PackageVersion Include="Microsoft.DotNet.VersionTools" Version="$(MicrosoftDotNetVersionToolsVersion)" />
     <PackageVersion Include="Microsoft.DotNet.Web.Authentication" Version="$(MicrosoftDotNetWebAuthenticationVersion)" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="6.0.1" />
-    <PackageVersion Include="Microsoft.Extensions.Configuration.Json" Version="6.0.0" />
+    <PackageVersion Include="Microsoft.Extensions.Configuration" Version="8.0.0" />
```
garath, line 60: I recognize that these updates are safe because they all implement netstandard2.0, but I'm wondering what motivates the bump in this PR?

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        /// <summary>
+        /// Check for local environment values to indicate you are running for Azure DevOps
+        /// SYSTEM_COLLECTIONURI is a default environment variable in Azure DevOps
+        /// </summary>
+        private bool RunningInAzureDevOps = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("SYSTEM_COLLECTIONURI"));
+
+        /// <summary>
+        /// Provides the ServiceTreeId set with global options
+        /// The ID is a goid and is set to Guid.Empty if not set
```
garath, line 27:
```suggestion
        /// The ID is a guid and is set to Guid.Empty if not set
```

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,94 @@
...
+        /// Base constructor for the ProjectBaseCommand class
+        /// </summary>
+        public ProjectBaseCommand(GlobalCommand globalCommand)
+        {
+            Verbosity = globalCommand.Verbosity;
+            Help = globalCommand.Help;
+        }
+
+        /// <summary>
+        /// Overides the GetOptions method from the base class to add a cusotm option for the ServiceTreeId
```
garath, line 41:
```suggestion
        /// Overrides the GetOptions method from the base class to add a custom option for the ServiceTreeId
```

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateAllCommand.cs
```diff
@@ -45,6 +47,9 @@ public override bool AreRequiredOptionsSet()
 
     public override async Task RunAsync(CancellationToken cancellationToken)
     {
+        // Provides a curtisy warning message if the ServiceTreeId option is set to a empty guid
```
garath, line 50:
```suggestion
        // Provides a courtesy warning message if the ServiceTreeId option is set to an empty guid
```

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            ControlPanelLogger.LogAudit(auditRecord);
+        }
+        internal static string GetLocalIPAddress()
+        {
+            var host = Dns.GetHostEntry(Dns.GetHostName());
+            var ipAddress = host.AddressList.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork);
+            if (ipAddress == null)
+            {
+                throw new Exception("No network adapters with an IPv4 address in the system!");
```
garath, line 105: No IPv6?

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/AzureKeyVault.cs
```diff
@@ -107,44 +110,106 @@ private static ImmutableDictionary<string, string> GetTags(global::Azure.Securit
...
+           operationResultMessage = e.Message;
+           throw;
+        }
+        finally
+        {
+            // Record an audit log for the secret update operation
+            _auditLogger.LogSecretUpdate(
+                credentialProvider: _tokenCredentialProvider,
+                secretName: name,
+                secretStoreType: "AzureKeyVault",
```
garath, line 148:
```suggestion
                secretStoreType: nameof(AzureKeyVault),
```
Compiler fanciness.

garath (changes requested):
Marking to hold at least until the questions about how CI will handle the new packaging are answered.

Directory.Packages.props
davfost: The only actual NuGet I wanted to add was OpenTelemetry.Audit.Geneva, but I believe it or one of it's dependencies required other updates to occur.

NuGet.config
davfost: Yes. I don't think we can merge (validly) until this is address as it build breaks at the NuGet restore step. So, it has to be fixed in some way in order to pass the PR validation build.

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
davfost: If I understand correctly I think you would be describing a Stub class which I think would be fine. However, the SecurityAutidLogger class is supposed to be non-volatile meaning it should not throw even if logging fails. So, if this does throw in any test it would mean that there is a problem with the class.

NOTES: 

SecurityAutidLogger is a wrapper class used to simplify calls to the OpenTelemetry.Audit.Geneva methods. This is because the OpenTelemetry.Audit.Geneva operation have heavy setup costs and are volatile. So, this wrapper reduces the setup costs of audit logging and ensure the logging operations are non-volatile.

davfost: Please confirm if you want me to make a stub for this class. My take is that creating a stub is an extra thing we would have to manage, and this class is supposed to be non-volatile so if it is throwing unexpectedly in test cases that is a real problem not a false positive. But I am fine making a stub if you prefer it that way.

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
davfost: I would like to have an offline conversation with you about commenting in general. I was thinking about trying to start a team wide discussion but I would like to get some individual views first.

In my view the first set of comments should exist because it is not reasonable to assume a new Dev will intuitively know 'why' injection instruction were added. Meaning what are the objects used for and why are they needed. Yes you can 'figure it out' if you go back though history looking for when the lines were added and then examine the other code changes (hoping that it was a small well scoped change) to determine what caused the injection to be added or do global search fining where the objects are used and figuring out what they do in those classes. However, that is an investigation delay that can be avoided if you have a comment.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateAllCommand.cs
davfost: The call to base.GetOptions().**AddRange(**

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
davfost: It is needed for the call to base.GetOptions().**AddRange**

Is there a reason we want to avoid using Microsoft.VisualStudio.Services.Common?

I think there may be another way to do this like getting the base object then calling Add individually for every new option.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: If service tree is empty 'this' class will suppress logging but the actual logging processes from OpenTelemetry.Audit.Geneva will not. If you provide a empty Guid it just pass it though as a string in my testing that did not result in an error just a log with a empty guid string for the service id.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: I am not sure if this is supported on Linux, which is one of the reasons to create a wrapper to make the process non-volatile. 

This process logs to an intermediate location (ETW was the recommended path) but that does not give you the Geneva Audit Logging.

In order to get the actual logging to Geneva to occur you have to have an Geneva agent running on the system where logs are written that periodically collects data and transfers it to a temporary storage location where another agent collects that data from storage and then moves it to a Kusto DB. 

These are the options available:

public enum AuditLogDestination
{
    SecurityEvent,
    ETW,
    UnixDomainSocket,
    TCP
}

I know the first 2 are supported in windows environments and I believe ETW is supported in azure environments which is why it is the recommended path.

**Do the either of the other types look familiar in terms of Linux support?**

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: I think so as we are not passing any 'secret' data in the logs that can be exposed. The identity information is the information  provided by the caller so they should already know it or have access to it. If there is a concern we can remove the ex.Message component to be extra safe but it will make debugging harder if we every have to investigate and issue where expected logs are missing.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
dougbu: > Is there a reason we want to avoid using Microsoft.VisualStudio.Services.Common?

probably not. suspect @mmitche was just making sure the `using` was truly needed

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
dougbu: `UnixDomainSocket` and `TCP` should definitely be available on Linux. might get hints about `ETW` support when looking at available collection agents are supported

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
dougbu: I prefer we leave historical comments out of the code though they sometimes make sense in commit descriptions. investigations that go back in history tend to be about blame or providing more descriptive information for a fix e.g., "this partially reverts {some commit} to avoid {some problem}". but the history doesn't usually help figuring the fix out in the first place. put another way `git log` and similar for history, not comments

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        /// Check for local environment values to indicate you are running for Azure DevOps
+        /// SYSTEM_COLLECTIONURI is a default environment variable in Azure DevOps
+        /// </summary>
+        private bool RunningInAzureDevOps = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("SYSTEM_COLLECTIONURI"));
+
+        /// <summary>
+        /// Provides the ServiceTreeId set with global options
+        /// The ID is a guid and is set to Guid.Empty if not set
+        /// </summary>
+        public Guid ServiceTreeId = Guid.Empty;
```
dougbu, line 29: why a publicly-writable field❓ seems much more normal to only expose `const` fields and properties

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        }
+
+        /// <summary>
+        /// Overrides the GetOptions method from the base class to add a custom option for the ServiceTreeId
+        /// </summary>
+        public override OptionSet GetOptions()
+        {
+            return new OptionSet()
+            {
+                {"servicetreeid=", "The service tree ID (must be a valid GUID id from aka.ms/servicetree)", id =>
```
dougbu, line 47: what enforces validity in Service Tree❓

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+using Mono.Options;
+
+namespace Microsoft.DncEng.SecretManager.Commands
+{
+    /// <summary>
+    /// This class is used to extend the CommandLineLib.GlobalCommand class to add a global options spacific to this project
+    /// </summary>
+    public class ProjectBaseCommand : GlobalCommand
+    {
+
```
dougbu, line 13: nit:
```suggestion
```

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,89 @@
...
+        {
+            return new OptionSet()
+            {
+                {"servicetreeid=", "The service tree ID (must be a valid GUID id from aka.ms/servicetree)", id =>
+                    {
+                        if (Guid.TryParse(id, out var guid))
+                        {
+                            ServiceTreeId = guid;
+                        }
+
```
dougbu, line 53: nit
```suggestion
```

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,50 @@
...
     public static Task<int> Main(string[] args)
     {
+        // The GlobalCommand must be pre-parsed before passing to the  ProjectBaseCommand object or the base global settings will be lost
+        var options = _globalCommand.GetOptions();
+        options.Parse(args);
+
+        // We then parse the ProjectBaseCommand to ensure we collect the service tree id at the start of the progress
+        // so it can be used for dependency ingjection
+        // The global option setings are passed to all other command objects that inhearit from the ProjectBaseCommand
+        var projectBaselCommand = new ProjectBaseCommand(_globalCommand);
```
dougbu, line 38:
```suggestion
        var projectBaseCommand = new ProjectBaseCommand(_globalCommand);
```
and update references to the variable

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,50 @@
...
+        options = projectBaselCommand.GetOptions();
+        options.Parse(args);
+        ServiceTreeId = projectBaselCommand?.ServiceTreeId ?? Guid.Empty;
```
dougbu, line 41: variable cannot be `null`
```suggestion
        ServiceTreeId = projectBaselCommand.ServiceTreeId ?? Guid.Empty;
```

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,50 @@
...
 namespace Microsoft.DncEng.SecretManager;
 
 public class Program : DependencyInjectedConsoleApp
 {
+    /// <summary>
+    /// Object stores global command setting as parsed from the command line at the main method
+    /// We mark this valud as protected so it can be accessed by processes that invoke the assembly outside of the command line
+    /// </summary>
+    protected static GlobalCommand _globalCommand = new GlobalCommand();
+
```
dougbu, line 21:
```suggestion
```

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: Copilot added IpV6 Support

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: One other note I will follow up with Dale on the DDFun team as I believe the code they deployed they did use ETW on Linux machines, but it required adding a 'mounting' action of some kind. (I know next to nothing about Linux, so I will need them to explain what they did.)

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
davfost: ServiceTreeId is a property value that must be access outside of the class so it can't be private or hidden, and it is not a const because it is alterable via command option settings passed in by the caller.

I have altered the code to use an private class variable to data storage and exposed the property as a 'get' only to make it read only outside of option setting operations.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
davfost: Nothing other than it being a 'Guid'. The audit logging process from does not verify that a ID is valid just that it is passed and the value meets the data type criteria.

If you pass an empty Guid or a valid Guid for a non-existent service tree ID it would still 'log' to Geneva successfully.

I don't think we have an additional obligation to make this process have to communicate with an external Service Tree API to ensure IDs are valid. 

If the 'service' does not use a correct ID then it will still report to Geneva but Geneva will not close any SFI KPI work items associated with the service so in that sense you are incentivized to use a valid and correct ID.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
garath: I wouldn't want to add this package just to get this extension, but since it's already required for the project you may as well get milage out of it.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,49 @@
...
     public static Task<int> Main(string[] args)
     {
+        // The GlobalCommand must be pre-parsed before passing to the  ProjectBaseCommand object or the base global settings will be lost
+        var options = _globalCommand.GetOptions();
+        options.Parse(args);
+
+        // We then parse the ProjectBaseCommand to ensure we collect the service tree id at the start of the progress
+        // so it can be used for dependency ingjection
+        // The global option setings are passed to all other command objects that inhearit from the ProjectBaseCommand
+        var projectBaseCommand = new ProjectBaseCommand(_globalCommand);
```
dougbu, line 37: again, looks very strange to inherit from `GlobalCommand` and pass a `GlobalCommand` instance into the constructor. also feels odd for regular commands such as `ValidateCommand` to inherit from `GlobalCommand`

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
```diff
@@ -8,16 +11,40 @@ namespace Microsoft.DncEng.SecretManager;
...
+    {
+        // Get a token from the crendential provider
+        var tokenRequestContext = new TokenRequestContext(new[] { "https://management.azure.com/.default" });
+        var token = _credential.Value.GetToken(tokenRequestContext, CancellationToken.None);
+
+        // Decode the JWT to get user identity information
+        var handler = new JwtSecurityTokenHandler();
+        var jsonToken = handler.ReadToken(token.Token) as JwtSecurityToken;
+        ApplicationId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "appid")?.Value ?? "Claim appid (Application Id) Not Found"; 
+        TenantId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "tid")?.Value ?? "Claim tid (Tenant Id) Not Found";
```
dougbu, line 48: should these expressions `throw` instead of setting the properties to invalid values❓

dougbu (commented):
I think the new comments contain a few more spelling errors

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
davfost: I will try to use the Moq lib to resolve this based on our offline conversation.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
garath: I do not think there is any reasonable way for this library to ensure it is a valid service tree ID. I think about this instead as being sort of "tagging" information or metadata that gets passed to Geneva, then Geneva happens to do something interesting with it.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: We believe that the ETW setting is most correct at this time, there are other options available which we can adept if needed in the future.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: I will remove the exception for now, if debugging becomes and issue we can come back to this.

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
davfost: I will remove the historical comment.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
dougbu: I'm fine w/ the level of validation implemented here. my main concern is the option's description implies something more. how about the following❓
```suggestion
                {"servicetreeid=", "The service tree ID (must be a non-empty GUID and should be a valid GUID id from https://aka.ms/servicetree)", id =>
```

dougbu (commented):
sorry, more spelling

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
```diff
@@ -7,9 +10,29 @@ public sealed class WrappedTokenProvider : ITokenCredentialProvider
...
+        {
+            // Get a token from the crendential provider
+            var tokenRequestContext = new TokenRequestContext(new[] { "https://management.azure.com/.default" });
+            var token = _tokenCredential.GetToken(tokenRequestContext, CancellationToken.None);
+
+            // Decode the JWT to get user identity information
+            var handler = new JwtSecurityTokenHandler();
+            var jsonToken = handler.ReadToken(token.Token) as JwtSecurityToken;
+            ApplicationId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "oid")?.Value ?? "Claim Oid Not Found";
+            TenantId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "tenant_id")?.Value ?? "Claim tenant_id Not Found";
```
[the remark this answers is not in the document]
davfost: There are a couple of reasons not to throw.

- We want this logging operation to be non-volatile no matter what. It is better not to log anything than have logging or logging setup cause our service to fail.
- The correctness of these value is not enforced by Geneva 
    - They force us to pass a value but do not verify it is a correct or valid Id just that a value exists
- Logs are still valuable even if we cannot get ever bit identifying data that would want so we would prefer to log something rather than nothing


Looking at this code again throw I think it should also be wrapped in a try catch block to ensure that even if the process of extracting clams value throws an exception the process does not cause the service to fail.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
davfost: I get what you mean about the description for input but 'should be' implies that something else is OK to use here.

It is not. The value is only meaningful if it is a valid service tree ID and specifically the service tree ID for your service. 

Any other Guid even one for a valid service tree ID for a service other than the one running the process would be invalid.

The value 'would' be accepted by Geneva and recorded but it is not correct and does not benefit the actual service that needs to report so it can close it's KPIs.

However, the way this option is being provided allows for service we do not own to also use this tool and generate a log that can go to Geneva.

We cannot predict what those possible IDs would be or how they will change in the future.

What we do know is that the ID is meant to be a Guid structure that comes from https://aka.ms/servicetree and relates to the specific service owned by the 'caller'.

I will change the comment to be something less directive so it does state must or should and instead just provides the link back to the location where valid IDs can be found.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,49 @@
...
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Http;
 
 namespace Microsoft.DncEng.SecretManager;
 
 public class Program : DependencyInjectedConsoleApp
 {
+    /// <summary>
+    /// Object stores global command setting as parsed from the command line at the main method
+    /// We mark this valud as protected so it can be accessed by processes that invoke the assembly outside of the command line
```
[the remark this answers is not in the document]
davfost: It is my understanding that this code is deployed in a nuget package and used directly as an assembly.

This means the package can be consumed and used as an assembly in a project. 

I don't know if it is or is not used that way, but I believe it is possible to be used that way.

If it is ever used that way, then you would need to set these 2 values directly. Assuming that they also leverage the protected method ConfigureServices to setup injection operations.

**_If that is considered an unsupported scenario, I would be fine with removing the comment and changing the protection levels to private._**

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
davfost: I agree but I am not familiar with Microsoft.DncEng.CommandLineLib options parsing logic and I was unable to find a different way to override global command or access them at this stage in the process.

Ideally I would be creating an override of the GlobalCommand object and adding the new ServiceTreeId property to it and then accessing the GlobalCommand inside of the other command types.

However, there seems to be some kind of high level injection override that prevents creating a custom 'GlobalCommand' object and forces the use the 'base' type.

The only examples I found for relaying 'global' data down to local commands required pre-parsing the global command in the 'main' method and passing the global setting into the other command types via constructor arguments.

Since I was not able to alter the override GlobalCommand object I instead created a base command type and then did the required pre-parsing of data to get the values needed before injection operations are carried out and passed the GlobalCommand data down to the other command types via the base command class I created. 

If someone has an example of a better way to do this, I am more than a happy to make a change.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
davfost: Similar to the other comment I mad this method non-volatile. We don't want this process to kill the service if it fails and we don't want to lose logging even these values are not found. The logs still have value even if not all identity values are found.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
dougbu: I'm not positive but looking at the code in dotnet/dnceng-shared, it feels like the global command isn't really intended to be extended. may make more sense for `ProjectBaseCommand` to inherit from `Command`.

note `DefaultConsole.ShouldWrite(VerbosityLevel)` exposes `GlobalCommand.Verbosity` indirectly and the `--help` command (all that `GlobalCommand` supports) is handled internally w/in the `CommandLineLib` package

davfost: DefaultConsole appears to be a Consol Writer tool that uses ICommandOptions from it's constructor to extract a 'GlobalCommand' object. 

It does not seem correct to leverage a console accessor object as a wrapper to access another object type that happens to be exposed inside that class. 

image

I need the consol object or the methods that DefaultConsole I just need to access to global command options so the global setting can actually be applied in sub commands.

However, I will see if I can leverage the same ICommandOptions interface in the same way DefaultConsole does to access the GlobalComman object and avoid creating it and parsing it directly in the Main method.

Even if that does not work or allow for removing the global object parsing in Main, I can still make ProjectBaseCommand inherit from Command, it only inherits from GlobalCommand to reduce the amount of redundant property settings.

dougbu: an option which applies to most of the available commands (e.g. does the service tree id add anything useful to `InfoCommand` or `TestCommand`❓) doesn't feel "global" to me. it may look awkward to have the same option in multiple `Command` subclasses but not as awkward as the current code.

separately, the parse ordering points seem less relevant to me. why does the service tree id option need to be parsed before anything else❓ is some logging done prior to the start of the `Command` execution❓ the code in `DependencyInjectedConsoleApp` creates the `ServiceCollection` within its `RunAsync(...)` method

davfost: You seem very concerned about this code and its implementation perhaps should have an offline conversation to address your concerns. We should also include Matt and Michael.

To be clear I have no passion one way or another about how the command parsing or injection logic is carried out as that has nothing directly to do with the objective of the SFI work.

The only reasons I am messing with these operations at all is because it was decided to pass in the Service Tree ID instead of hard coding it to one of our internal service tree id value since this process can be used by other teams for other services.

The ultimate goal is to be able to ensure any command that needs to perform audit logging can instantiate a logging object safely without producing mystery run time failure if other logging needs to be added in the future.

I will look to setup a meeting between you me and Matt to review and discuss the specifics of the implementation.

davfost: - Service Tree ID need to be parsed from the command line in order to passed to the Audit Logger object (design requirement) 
- The Audit Logger Object is part of this project space to it is defined in the ConfigureServices
- The call to ConfigureServices happens before the Argument Parser logic actually parses arguments.

The reason service tree needs to be manually parsed is because the audit logger is instantiated by the injection logic before the 'normal' argument parsing occurs. 

I have found no way to circumvent these order of operation issues so far.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+                options.Destination = AuditLogDestination.ETW;
+                options.ServiceId = serviceTreeId;
+                // If the service ID is a empty guid we should suppress audit logging
+                if (serviceTreeId == Guid.Empty)
+                {
+                    SuppressAuditLogging = true;
+                }
+            });
+
+            ControlPlaneLogger = auditFactory.CreateControlPlaneLogger();
```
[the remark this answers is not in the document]
davfost: In the most recent changes based on feedback from security advisors we will no longer suppress these logs for any reason. Logging will always occur with as much valid data as possible.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,92 @@
...
+                            throw new ArgumentException($"Failed to parse a valid Guid value from ServiceTreeId value '{id}'!");
+                        }
+                    }
+                } 
+            };
+        }
+
+        /// <summary>
+        /// Provides a non-volitie warning message if the ServiceTreeId option is set to a empty guid value and argments have been parsed
+        internal void ValidateServiceTreeIdOption()
```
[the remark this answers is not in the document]
davfost: Renamed

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/SynchronizeCommand.cs
```diff
@@ -42,19 +45,22 @@ public override List<string> HandlePositionalArguments(List<string> args)
...
             {"verify-only", "Does not rotate any secrets, instead, produces an error for every secret that needs to be rotated.", v => _verifyOnly = !string.IsNullOrEmpty(v)},
-        };
+        });
     }
 
     public override async Task RunAsync(CancellationToken cancellationToken)
     {
         try
         {
+            // Provides a curtisy warning message if the ServiceTreeId option is set to a empty guid
```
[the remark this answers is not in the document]
davfost: Renamed.

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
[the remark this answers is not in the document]
davfost: Code and comments were removed in latest implementation

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -4,22 +4,49 @@
...
+    /// </summary>
+    protected static GlobalCommand _globalCommand = new GlobalCommand();
```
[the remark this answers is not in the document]
davfost: Global command are no longer used or inherited from in latest implementation

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
davfost: Global Commands were removed in latest implementation.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
davfost: Based on the latest security advisor responses this logging will always occur in a non-volatile way. Meaning we always log with as much valid information as possible and any values we cannot extract are set to some default value.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// <summary>
+        /// Provides the ServiceTreeId set with global options
+        /// The ID is a guid and is set to Guid.Empty if not set
+        /// </summary>
+        private Guid _ServiceTreeId = Guid.Empty;
+
+        /// <summary>
+        /// Provides read only access to the _ServiceTreeId
+        /// </summary>
+        public Guid ServiceTreeId { get { return _ServiceTreeId; } }
```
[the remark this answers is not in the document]
davfost: The class has been replaced with CommonIdentityCommand.

This is the property setting now:

public Guid ServiceTreeId { get; private set; } = new Guid("8835b1f3-0d22-4e28-bae0-65da04655ed4");

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+namespace Microsoft.DncEng.SecretManager
+{
+    /// <summary>
+    /// SecurityAuditLogger is a class that is used to log security audit events to the security event log and Geneva
+    /// </summary>
+    public class SecurityAuditLogger
+    {
+        private ILogger ControlPlaneLogger;
+
+        private bool SuppressAuditLogging = false;
```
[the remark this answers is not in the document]
davfost: Setting was removed because we no longer suppress logging for any reason based on responses from security advisors.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,125 @@
...
+            // which are not static and not defined by the service
+            // So we are specifying what we belive the minmal acces level
+            // would be requried for this operation to be successful
+            auditRecord.AddCallerAccessLevel("Writer");
+            auditRecord.AddTargetResource(secretStoreType, secretLocation);
+            auditRecord.OperationResultDescription = (!string.IsNullOrWhiteSpace(resultMessage)) ? $"{resultMessage}" : $"'{operationName}' : '{result}'";
+
+            ControlPlaneLogger.LogAudit(auditRecord);
+        }
+        internal static string GetLocalIPAddress()
```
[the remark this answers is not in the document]
davfost: Typically it would be so they can be used in Unit Tests, since they are not I changed the access level to private.

NuGet.config
davfost: HI @garath 

Ok it looks like we have a potential solution for this issue using #IF statement to conditionalize the use of the OpenTelemetry.Audit.Geneva package only for 'Internal' builds.

Please take a look at the use of #IF blocks in 

AzureKeyVault.cs
SecurityAuditLogger.cs

Also take a look at the conditional specified in the following files:

Directory.Build.props
Microsoft.DncEng.SecretManager.csprog

NOTE: This solution would require us to add the following parameter to donet build arguments for internal builds:

-p:DefineConstants="INTERNAL"

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/GitHubHookController.cs
```diff
@@ -102,7 +102,8 @@ public async Task<IActionResult> PullRequestReviewComment()
         string uri = payload.Comment.HtmlUrl;
         string username = payload.Comment.User.Login;
         DateTimeOffset date = payload.Comment.UpdatedAt;
-        using IDisposable scope = _logger.BeginScope("Handling pull request {repo}#{prNumber}", repo, number);
+        using IDisposable scope = _logger.BeginScope("Handling pull request {repo}#{prNumber}", repo, number)
```
davfost, line 105: Review the changes to the using statements in this controller file with @garath 

The changes were added to address the build failures related to BeginScope returning a 'nullable' typed IDisposible object. Those errors appear to be treated as warning in local builds.

davfost:
> Marking to hold at least until the questions about how CI will handle the new packaging are answered.

Please see my response to the comment on this issue that discusses the solution being attempted.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
davfost: This file no longer exists but the latest version also has this property and it is public. The service tree ID is used in the constructor of another class during dependency injection operations so it needs to be public to be used there.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ProjectBaseCommand.cs
davfost: Nothing enforces the use of a 'real' service tree id including the tools provided by the Geneiva auditing team. Since they don't require a 'real' id we do not require a 'real' id. We do however require that the ID be the same data type that is expected for service tree IDs.

src/SecretManager/Microsoft.DncEng.SecretManager.Tests/SynchronizeCommandTests.cs
davfost: Updated with Mock objects

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/GitHubHookController.cs
garath: Hm, does changing the return type to `IDisposable?` work? That seems a bit more straightforward.

davfost: Yes that worked.

Directory.Build.props
```diff
@@ -57,4 +57,16 @@
...
+  <!--  
+  The following Target setting is used to conditionally add package source to the nuget.config for this solution.
+  We do this so we can add restricted access package sources to internal builds that cannot be access in public builds.
+  To set this build condition add the following setting  <-p:DefineConstants="INTERNAL"> to a 'dotnet build' command.
+  -->
+  <Target Name="UpdateNuGetConfig" AfterTargets="BeforeBuild">
+    <ItemGroup>
+      <NuGetConfigFiles Include="NuGet.config" />
+    </ItemGroup>
+    <Exec Condition=" '$(DefineConstants)' != '' AND '$(DefineConstants.Contains(INTERNAL))' " Command="dotnet nuget add source -n AzureGenevaMonitoring https://msblox.pkgs.visualstudio.com/_packaging/AzureGenevaMonitoring/nuget/v3/index.json --configfile %(NuGetConfigFiles.FullPath)" />
```
garath, line 70: This fails for me when I run dotnet build -p:DefineConstants="INTERNAL" at both the Solution level and at the SecretManager project level. What am I missing?

garath: Related, can't the reference to the feed stay in NuGet.config? I thought it would just be skipped and as long as the desired packages were found in other feeds no error would be produced.

davfost: NuGet.config was not altered, there is logic in Directory.Build.Props that will inject the new source, but only if the new Constant valid is defined.

So if it is trying to load the nuget it should mean the property was successfully passed in during the build.

garath: If the feed can't simply stay in the NuGet.Config, perhaps Package Source Mapping will help by making it clear that NuGet shouldn't even try to check the feed unless it's looking for that package.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/AzureKeyVault.cs
```diff
@@ -1,13 +1,16 @@
...
 using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
+using Azure;
+using Azure.Security.KeyVault.Keys;
+using Azure.Security.KeyVault.Secrets;
+using JetBrains.Annotations;
+using Microsoft.DncEng.CommandLineLib;
+#if INTERNAL
+using OpenTelemetry.Audit.Geneva;
```
garath, line 12: I don't love that the ifdef'ing leaks to other unrelated files. Have you considered adding whatever this file needs from the Geneva namespace into your abstraction?

garath:
Note that you'll also need to change how this project gets published. The internal build of SecretManager will need to use the new "INTERNAL" flag and publish to an internal-only feed. Then, all consumers will need to be updated to pull that internal instance from the internal-only feed.

src/SecretManager/Microsoft.DncEng.SecretManager/StorageTypes/AzureKeyVault.cs
davfost: Alter code so only 1 file should need the IF condition.

NuGet.config
davfost: I have now tried numerous methods to get passed the NuGet feed failures form the PR build that occur when the Geneva Audit feed is added to the nuget.config file. 

All conditional overrides specified in props or proj files either only work for one of the 2 scenarios (public/internal) or result in some kind of timing issue where the first build attempt for 'internal' operation fail.

The most successful conditional override in a props file involved adding a setting to edit the nuget.config file dynamically,  but that opens the door for unintentional code modification if a dev is testing 'internal' code.

We attempted to use nuget.config setting to specify explicit source mappings, but the build | publish task still attempt to access the Genva Feed in the 'public' build process which results in a failure.

**At this time I have removed all conditional setting and am planning to rely on 'build time' argument value to specify an alternat nuget.config file for 'internal' builds.**

The internal command would look like this:

dotnet build ./dnceng.sln -p:DefineConstants="INTERNAL" --configfile "nuget.internal.config"

Directory.Build.props
davfost: Package Source Mapping did not work.

I have tried several other methods to conditionally set Source Feed values from prog and props files but they either end up only working in one environment (pubic VS internal) or have timing issue that make build outcomes unpredictable. 

**At this time I am falling back on setting an override in the build command:

Example:**
dotnet build ./dnceng.sln -p:DefineConstants="INTERNAL" --configfile "nuget.internal.config"

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/CommonIdentityCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+    /// <summary>
+    /// This class is used to extend the CommandLineLib.Command class to provide common identity value options which can be used by other commands via inheritance
+    /// </summary>
+    public class CommonIdentityCommand : CommandLineLib.Command
+    {
+        /// <summary>
+        /// Check for local environment values to indicate you are running for Azure DevOps
+        /// SYSTEM_COLLECTIONURI is a default environment variable in Azure DevOps
+        /// </summary>
+        private bool RunningInAzureDevOps = !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("SYSTEM_COLLECTIONURI"));
```
garath, line 19: I would prefer to see things like this come via a configuration object or some other specific configuration path. I probably wouldn't hold the PR for it, though.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/CommonIdentityCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// <summary>
+        /// Provides the ServiceTreeId set with global options
+        /// The ID is a guid and is set to the Helix service tree ID by default
+        /// </summary>
+        public Guid ServiceTreeId { get; private set; } = new Guid("8835b1f3-0d22-4e28-bae0-65da04655ed4");
```
garath, line 25: I don't think using the Helix ID is a good default. Is there some way to make it a required configuration that gets piped through from the top-level app configuration?

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/CommonIdentityCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+                        }
+                    }
+                }
+            });
+            return options;
+        }
+
+        /// <summary>
+        /// Provides a non-volatile warning message if the ServiceTreeId option is set to a empty guid value
+        internal void WarnIfServiceTreeIdIsSetToEmptyGuid()
```
garath, line 63: Given that the "non-INTERNAL" version of the app is expected to be used for a while longer, it would be good to silence warnings like this.

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
```diff
@@ -7,9 +10,37 @@ public sealed class WrappedTokenProvider : ITokenCredentialProvider
...
+                var tokenRequestContext = new TokenRequestContext(new[] { "https://management.azure.com/.default" });
+                var token = _tokenCredential.GetToken(tokenRequestContext, CancellationToken.None);
+
+                // Decode the JWT to get user identity information
+                var handler = new JwtSecurityTokenHandler();
+                var jsonToken = handler.ReadToken(token.Token) as JwtSecurityToken;
+                ApplicationId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "oid")?.Value ?? "Claim Oid Not Found";
+                TenantId = jsonToken?.Claims?.FirstOrDefault(claim => claim.Type == "tenant_id")?.Value ?? "Claim tenant_id Not Found";
+            }
+            catch
```
garath, line 39: Catch-everything seems heavy-handed. Specifically, this will hide a failure to actually get a token (e.g. permissions problems) and it seems that should be bubbled up to the rest of the app.

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
```diff
@@ -7,9 +10,37 @@ public sealed class WrappedTokenProvider : ITokenCredentialProvider
...
+            {
+                ApplicationId = "Failed To Read Claims Data: Oid";
```
garath, line 41: I prefer a string meant to hold "Application ID" not do double-duty with ".. or an error message". I suggest marking the field as nullable and letting the null value mean it's not available.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
garath, line 1: Same comments here as in the WrappedTokenProvider class

src/SecretManager/Microsoft.DncEng.SecretManager/Program.cs
```diff
@@ -1,25 +1,45 @@
...
     }
 
     protected override void ConfigureServices(IServiceCollection services)
     {
+        //// Dependency injection instruction needed to support properties used for Geneva Logging operations
+        services.AddSingleton<CommonIdentityCommand>();
+        services.AddSingleton(serviceProvider =>
+        {
+            var baseCommand = serviceProvider.GetRequiredService<CommonIdentityCommand>();
+            // We pre-pars command argument here to overcome a order of operations issue where
```
garath, line 33:
```suggestion
            // We pre-parse command argument here to overcome a order of operations issue where 
```

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: Then the process is no longer 'non-volatile' it is now a requirement that valid credentials can be extracted using this code. 

That is not desirable here. We don't want this logging operation or any of the required data gathering efforts to cause the actual process to fail.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/CommonIdentityCommand.cs
davfost: The consumer can already override this value using the command line, why is having another 'default value' override valuable?

The reason I picked the Helix ID here is so that we don't have to do 2 more rollouts to clear the SFI KPI for this issue. The helix service is the only one that is currently flagged so getting some kind of log to report for that service is the goal.

**_If the idea her is just not liking a hard coded string for a Guid ID to exist in code, I would say we can do the 3rd deployment to explicitly set the ID in processing calls and to change this to a 'empty' Guid by default after the KPI is cleared._**

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: I remove the guidance message from the ID values on a failure and replaced that with comments explaining why we want this process to be non-volatile.

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/CommonIdentityCommand.cs
davfost: Based on our offline conversation, I will remove this warning.

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: Based on our offline conversation, I will remove the notes in the ID values when claims are not found and then resolve this item.

garath (dismissed):
All my questions for this iteration are answered!

garath (approved):

PR #3881, opened 2024-08-26
Remove dotnetengstatus connection strings

mmitche:
Not in use.

PR #4425, opened 2024-11-08
Remove unused secret dotnet-status-storage-account

chcosta:
Related issue: https://github.com/dotnet/dnceng/issues/4426

Fixes https://dev.azure.com/dnceng/internal/_workitems/edit/7126

AFAICT, this secret is no longer used and the app has been migrated to MI where needed.  After this change merges and secret manager is no longer attempting to rotate the value, I will get the KPI off the radar.


missymessa (approved):
LGTM shipit

@mmitche made the same changes here, but you're beating him to it: https://github.com/dotnet/dnceng/pull/3881/files

PR #4651, opened 2024-12-10
[Rollout] Production rollout, 2024-12-11

v-zheyan:
Main build: https://dnceng.visualstudio.com/internal/_build/results?buildId=2596198&view=results
Rollout issue: https://github.com/dotnet/dnceng/issues/4644


riarenas (approved):
I'm not sure but it seems like the diff might be confused. The code looks the same on both branches.

davfost:
It was decided that we would skip rollouts this week.

Issue #4658, opened 2024-12-11
Decommission Rollout Scorer


missymessa:
- [x] Remove Rollout Scorer deployment from dotnet/dnceng (this will alleviate the errors we're seeing when we do deployments)
- [x] Remove any code that does the markdown file generation
- [x] Remove Rollout Scorer resources from Azure (e.g., the Azure function, et cetera)
- [x] Keep anything that records metrics (e.g., when rollouts start and end)
- [x] Deprioritize SFI issues around Rollout Scorer resources. Review those issues and see what resources we can clean up (i.e., delete) from what we don't need anymore (because they were used for the Azure Function/service/markdown file generation)
- [ ] Re-priortize the replacement issue: https://github.com/dotnet/dnceng/issues/2373

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
Decommissioning Rollout Scorer as it's upkeep is costing more than it's worth.

missymessa:
Resource Groups to delete: 
- rollout-scorecards (Dotnet Engineering services)
- rollout-scorecards-staging (Dotnet Engineering services)
- rollout-scorer-prod (Helix)
- rollout-scorer-staging (HelixStaging)

missymessa:
Resource groups deleted.

missymessa:
We'll clean up any other SFI issues related to Rollout scorer as we see them. Closing this issue.

PR #4676, opened 2024-12-16
Remove RolloutScorer Azure Function from solution

missymessa:
https://github.com/dotnet/dnceng/issues/4658

Retains the CLI and library functionality that does the calculations and can generate a scorecard locally, but removes the Azure Function from the code base.


garath:
This doesn't remove the actual files on disk. What's the plan for those?

missymessa:
> This doesn't remove the actual files on disk. What's the plan for those?

There's an item in the linked issue to do that. I want to remove it from deployments/code base so it doesn't get added back to Azure accidentally :)

garath (approved):

garath:
> There's an item in the linked issue to do that. I want to remove it from deployments/code base so it doesn't get added back to Azure accidentally :)

Is that "Remove any code that does the markdown file generation" item? I'm nervous about having dotnet code in the repo that's no longer connected to a solution (it's super easy to overlook and forget about) but as long as the code deletion happening soon then it shouldn't be a problem.

missymessa:
> > There's an item in the linked issue to do that. I want to remove it from deployments/code base so it doesn't get added back to Azure accidentally :)
> 
> Is that "Remove any code that does the markdown file generation" item? I'm nervous about having dotnet code in the repo that's no longer connected to a solution (it's super easy to overlook and forget about) but as long as the code deletion happening soon then it shouldn't be a problem.

No, it should be covered under: "Remove Rollout Scorer resources from Azure (e.g., the Azure function, et cetera)"

I _expect_ anything Rollout Scorer related will be deleted (from dotnet/dnceng, specifically), but I wanted to make sure I was specific in this instance, as we'll want to keep the Azure table that logs the deployment times (so that new reporting can reference that data), so I didn't want resources related to that to be deleted.

PR #4750, opened 2025-01-08
Add Audit Logging To SecreteManager.exe attempt 2

davfost:
Add Audit Logging To SecreteManager.exe

- Adds the OpenTelemetry.Audit.Geneva package and dependencies
- Following logging recommendations from Geneva team
- Added non-volatile logging operations to secret update and create operations
- Adds new command line options for setting the target service ID to used when performing audit logging
- Corrects issue causing premature credential creation in the SecretManagerCredentialProvider class.

Part Of GitHub Issue https://github.com/dotnet/arcade/issues/13217


NuGet.Internal.config
```diff
@@ -0,0 +1,11 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
+    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
+    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
+    <add key="AzureGenevaMonitoring" value="https://msblox.pkgs.visualstudio.com/_packaging/AzureGenevaMonitoring/nuget/v3/index.json" />
```
mmitche, line 8: Eck...How do we switch over to using this when building internally?

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
```diff
@@ -7,9 +10,37 @@ public sealed class WrappedTokenProvider : ITokenCredentialProvider
...
+            SetCredentialIdentityValues();
+        }
+
+        /// <inheritdoc/>
+        internal void SetCredentialIdentityValues()
+        {
+            try
+            {
+                // Get a token from the credential provider
+                var tokenRequestContext = new TokenRequestContext(new[] { "https://management.azure.com/.default" });
```
chcosta, line 30: Is this doing actual authentication for tests?  I'd expect this to be a mocked request.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretManagerCredentialProvider.cs
```diff
@@ -8,16 +11,53 @@ namespace Microsoft.DncEng.SecretManager;
...
+    public string ApplicationId { get; internal set; }
+    /// <inheritdoc/>
+    public string TenantId { get; internal set; }
+
+    public SecretManagerCredentialProvider()
+    {
+    }
+
     // Expect AzureCliCredential for CI and local dev environments. 
-    // Use InteractiveBrowserCredential as a fallback for local dev environments.
```
chcosta, line 12: nit: I think `fallback` is the correct word here

NuGet.Internal.config
davfost: That is not in this PR.
There will be a second PR to enable 'interna' build switches and change where nuggets get deployed. (Those 2 updates need to happen in the same PR to ensure internal assemblies are not exposed in a public feed)
I have not determined how to change the feed we publish to yet.
However, I believe to apply the internal switch we will update the azure-pipelines.yml file to extend the '_InternalBuildArgs' varable to include the following parameters in the string <-p:DefineConstants="INTERNAL" --configfile "nuget.internal.config">

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: This only occurs if a valid token was pass into the process and it is designed to be non-volatile meaning in a the token is null or not valid the process will not fail and will exist clean.

However, I think that also means we can safely mock this operation for unit test so I will look into doing so.

davfost: This class is sealed so it cannot be directly mocked. Instead, I created a wrapper class in the unit test that ensure the existing test that use the WrappedTokenProvider class retain that functionality but it overrides the SetCredentialIdentityValues method to be a no-op.

garath: This class is used in a live test (notice the namespace is "ScenarioTest" which is what we've usually called thing that run against actual services). I don't think it needs to be mocked, just updated to implement the new interface methods.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
```diff
@@ -0,0 +1,191 @@
...
+
+        private void LogSecretAction(OperationType operationType, string operationName, ITokenCredentialProvider credentialProvider, string secretName, string secretStoreType, string secretLocation, SecretManagerOperationResult result, string resultMessage)
+        {
+            // The token application id of the client running the assembly.
+            // NOTE: The user identity here should be something 'dynamic'.
+            // If you are hard coding this value you should question if this Audit Log is useful
+            // as it is likely redundant to lower level permission change logging that is already occurring.
+            var user = credentialProvider.ApplicationId;
+            // Get the tenant ID that provided the token for the credential provider.
+            var tenantId = credentialProvider.TenantId;
```
garath, line 83: This value (and of `user` above) might be null. Are the usages below going to be okay wiht that?

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
```diff
@@ -14,19 +15,19 @@ public class ValidateCommand : Command
     private string _baseSettingsFile;
     private string _envSettingsFile;
 
-    public ValidateCommand(SettingsFileValidator settingsFileValidator)
+    public ValidateCommand(IConsole console, SettingsFileValidator settingsFileValidator)
```
garath, line 18: I don't see `IConsole` being used here, why add?

garath:
I believe you mentioned to me that you did, but just to confirm... you did test the scenario that caused the original change to be reverted?

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: I am not sure what the ask is here. Are you saying that the change to create a Test object should be revered so it uses the 'real' code in this test or is there something else that should be done here?

src/SecretManager/Microsoft.DncEng.SecretManager/Commands/ValidateCommand.cs
davfost: It was used in the previous implementation to pass down the common command base object, I will remove it now that that class has been removed.

src/SecretManager/Microsoft.DncEng.SecretManager/SecurityAuditLogger.cs
davfost: Yes these values are 'desired' for audit logging but will not 'break' logging if they are null.

davfost:
> I believe you mentioned to me that you did, but just to confirm... you did test the scenario that caused the original change to be reverted?

Yes, I re-run all unit tests and local validation tests including the test Ricard provided in the last roll back issue before checking in changes so I re-ran all test for the latest implementation and ran them again after making updates in the PR for comments.

When I make the other minor changes above I will re-run all the tests again to be sure nothing breaks.

src/SecretManager/Microsoft.DncEng.SecretManager.ScenarioTests/WrappedTokenProvider.cs
davfost: Per our offline conversation, I am rolling back the change to add a Test Object as the use of a real object with real credentials was intentional.

garath (approved):

PR #4878, opened 2025-02-01
Merge Production => main

davfost:
Main build: https://dnceng.visualstudio.com/internal/_build/results?buildId=2625115&view=results
Rollout issue: https://github.com/dotnet/dnceng/issues/4839

**NOTE: There is an unexpected change in this PR but it looks like the merge from production to main has not been done in a while.**


src/DotNet.Status.Web/DotNet.Status.Web/.config/settings.json
```diff
@@ -124,12 +124,6 @@
           "Branches": [ "main" ],
           "IssuesId": "dotnet-dnceng-first-responder"
         },
-        {
```
riarenas, line 127: Weird. We don't want to delete this in main, and we want this to be in production during the next deployment.

davfost: Ok I will abandon this PR since this is an expected and unwanted change.

davfost:
Abandoning merge back PR because it contains an unexpected and unwanted change to main.

PR #5025, opened 2025-02-18
Upload app package using assigned account

f-alizada:
Related issue:
https://github.com/dotnet/dnceng/issues/5026

Use the connected storage account during the CI run. 
Current approach uses the account key.


missymessa (approved):
LGTM shipit

PR #5062, opened 2025-02-21
Pcs git push alert

dkurepa:
https://github.com/dotnet/arcade-services/issues/4199

### DIFF
--- a/src/Monitoring/Monitoring.ArcadeServices/dashboard/arcade-services/arcadeAvailability.dashboard.json
+++ b/src/Monitoring/Monitoring.ArcadeServices/dashboard/arcade-services/arcadeAvailability.dashboard.json
@@ -149,7 +149,7 @@
         {
           "azureLogAnalytics": {
             "query": "customEvents\r\n| where name == \"GitPush\" and $__timeFilter(timestamp)\r\n| extend Success = tobool(customDimensions[\"Success\"])\r\n| summarize SuccessRate=100*countif(Success == true)/count() by bin(timestamp, $__interval)\r\n| order by timestamp asc",
-            "resource": "/subscriptions/fbd6122a-9ad3-42e4-976e-bccb82486856/resourceGroups/product-construction-service/providers/Microsoft.Insights/components/product-construction-service-ai-prod"
+            "resource": "[parameter(product-construction-service-appinsights-resourcepath)]"
           },
           "azureMonitor": {
             "dimensionFilters": [],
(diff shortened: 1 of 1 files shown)


adamzip (approved):

premun (changes requested):
I checked Grafana and a couple of things seem wrong

1. I think the panel needs to be moved inside the PCS row
2. We need to make sure that when there's no pushing, we don't alert:
    image
3. The condition seems to say "alert when success rate above 80%" - it should be below?
4. There's no notification channel
5. There's no `NotificationId` tag (check other alerts and make a unique one)

dkurepa:
> I checked Grafana and a couple of things seem wrong
> 
> 1. I think the panel needs to be moved inside the PCS row
> 2. We need to make sure that when there's no pushing, we don't alert:
>    image
> 3. The condition seems to say "alert when success rate above 80%" - it should be below?
> 4. There's no notification channel
> 5. There's no `NotificationId` tag (check other alerts and make a unique one)
> 6. The chart should have a min of 0 and max of 100 if it's %
> 7. The unit should be %

thanks, my bad for not checking these.. I moved the panel into the PCS row by hand, but I don't see any changes of the json config that would indicate that change, so not too sure if I it will remain there after deployment

dkurepa:
hmm, the notificationId tag also didn't stick

premun:
@dkurepa you need to click "Add tag" after (so a new empty row appears)

premun:
@dkurepa please move it to the bottom and keep it together with the git clone graph.

Usually I try to keep related panels below each other so that the charts can be visually compared (e.g. exceptions vs failed work items often share spikes).

premun:
@dkurepa can you pull the JSON again? I added the missing alert channel:
image


(I also added the alert threshold to the chart)

premun:
Can you check why it fired just now?
https://github.com/dotnet/dnceng/issues/5073

dkurepa:
This alert seems to be measuring when we have less than 20 items in 5 min, which happened. It's meant to check if we completely stop processing items, but maybe 20 is too high

dkurepa:
The alert message is wrong, will correct it in this PR

premun:
> The alert message is wrong, will correct it in this PR

Thanks

dkurepa:
Also, we're tagged twice in the description 
image
think this is so we also get tagged on every comment left by the bot, but we should get an email regardless since we've been mentioned in the issue?

dkurepa:
> think this is so we also get tagged on every comment left by the bot, but we should get an email regardless since we've been mentioned in the issue?
we don't get emails about updates, so maybe we should leave these there

premun:
I think it's not a big deal we're tagged twice.

premun (approved):

PR #5192, opened 2025-03-10
refactoring for nethelix-prod-kusto-ad-application and nethelix-staging-kusto-ad-application

engyebrahim:
https://dnceng.visualstudio.com/internal/_workitems/edit/7442/

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Pages/_Layout.cshtml
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Pages/_Layout.cshtml
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html>
+<html lang="en">
 
 <head>
     <meta charset="utf-8" />
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Pages/_Layout.cshtml
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Pages/_Layout.cshtml
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html lang="en">
+<html>
 
 <head>
     <meta charset="utf-8" />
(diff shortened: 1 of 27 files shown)


src/Telemetry/AzureDevOpsTimeline/KustoTimelineTelemetryRepository.cs
```diff
@@ -25,21 +26,16 @@ public sealed class KustoTimelineTelemetryRepository : ITimelineTelemetryReposit
     private readonly ICslQueryProvider _query;
     private readonly string _database;
 
-    public KustoTimelineTelemetryRepository(ILogger<KustoTimelineTelemetryRepository> logger, IOptionsSnapshot<KustoTimelineTelemetryOptions> options)
+    public KustoTimelineTelemetryRepository(ILogger<KustoTimelineTelemetryRepository> logger, IOptionsSnapshot<KustoOptions> options)
     {
         _logger = logger;
 
-        // removing the IngestConnectionString was a default setup in local debugging
```
garath, line 32: Will removing this break local development? Or is the `KustoIgnestFactory` below smart enough to "just do nothing" if there is no Managed Identity ID?

src/DotNet.Status.Web/DotNet.Status.Web/Pages/_Layout.cshtml
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 
-<html lang="en">
```
garath, line 3: Why remove this?

src/Telemetry/AzureDevOpsTimeline/.config/settings.Development.json
```diff
@@ -7,6 +7,8 @@
   "Interval": "0:30:00",
   "BuildBatchSize": 1000,
   "KustoTimelineTelemetry": {
-    "IngestConnectionString": ""
+    "KustoClusterUri": "https://engdata.westus2.kusto.windows.net",
```
garath, line 10: This is technically a change in behavior. Is there a need to change it?

engyebrahim: I updated the condition to check the managed identity instead, so in the code it's the same as the other settings

garath (approved):

Issue #5339, opened 2025-04-01
Automate `dn-dependabot-account-redmond` updates


dougbu:
when doing #5300, I made a local change to update the `dn-dependabot-account-redmond` settings automatically going forward. this PR is about committing that change in the dotnet/dnceng repo, together w/ similar settings for other Redmond bot accounts

### Release Note Category

- [ ] Feature changes/additions
- [ ] Bug fixes
- [X] Internal Infrastructure Improvements

### Release Note Description

made an internal change to automate future `dn-dependabot-account-redmond` updates

dougbu:
#5340 completed this work. updates have been rolled out

PR #5340, opened 2025-04-01
Automate `dn-dependabot-account-redmond` updates

dougbu:
- see https://github.com/dotnet/dnceng/issues/5339


dougbu:
could also do this in dotnet/arcade repo, where the value is referenced. however all (all two) other `domain-account` settings are in the same file I'm updating here and dotnet/arcade does not currently update HelixKV

riarenas (approved):

garath (approved):

PR #5341, opened 2025-04-02
Update `domain-account` secret type handling

dougbu:
- see https://github.com/dotnet/dnceng/issues/2365
- require (and document) use of CoreIdentity for password updates

### DIFF
--- a/src/SecretManager/Microsoft.DncEng.SecretManager/Readme.md
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager/Readme.md
@@ -45,13 +45,13 @@ secrets:
       blob: keys.xml
       container: dp
       permissions: racwd
-  
+
   execution-storage-connection:
     type: azure-storage-connection-string
     parameters:
       subscription: <a guid>
       account: helixexecution
-  
+
   execution-storage-container-sas:
     type: azure-storage-container-sas-uri
     parameters:
@@ -219,7 +219,7 @@ This is the most common option, and is used to store secrets in an Azure Key Vau
 
 This location represents service connections in an Azure DevOps instance. This is designed to support strictly support connections that require Azure DevOps PATs (secret type `azure-devops-access-token`) to operate (all other service connections should be using SFI-compliant patterns).
 
-Note that it is not possible to read secrets generated for this storage location. 
+Note that it is not possible to read secrets generated for this storage location.
 
 ## Secret Types
 Each secret in the manifest has a specified type, and these types determine what parameters are required, and what ends up in the secret store. Some secret types require human interaction to generate and/or rotate. For such secrets the tool will prompt the user or fail if in a build context. Some secret types also produce multiple values. These values are all stored with additional suffixes in the secret store. Each type is documented below.
@@ -336,7 +336,7 @@ type: kusto-connection-string
 parameters:
   dataSource: the DataSource in the connection string
   initialCatalog: the InitialCatalog in the connection string
-  additionalParameters: and extra 
+  additionalParameters: and extra
   adApplication: SecretReference to the ad-application used for authentication
 ```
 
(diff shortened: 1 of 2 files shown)


src/SecretManager/Microsoft.DncEng.SecretManager/Readme.md
```diff
@@ -45,13 +45,13 @@ secrets:
       blob: keys.xml
       container: dp
       permissions: racwd
-  
+
```
dougbu, line 48: recommend ignoring whitespace when reviewing this

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
```diff
@@ -30,24 +31,25 @@ protected override async Task<SecretData> RotateValue(Parameters parameters, Rot
...
+4. Copy the generated password to the clipboard before doing anything else. Paste it when requested in step 5.
+5. Run 'secret-manager --force-secret={parameters.AccountName}' to update the key vault with the new password.");
 
         if (!string.IsNullOrWhiteSpace(parameters.Description))
             _console.WriteLine($"Additional information: {parameters.Description}");
 
-        string newPassword = await _console.PromptAsync($"Enter a new password or press enter to use a generated password {generatedPassword} : ");
-        if (string.IsNullOrWhiteSpace(newPassword))
-            newPassword = generatedPassword;
+        string newPassword = await _console.PromptAndValidateAsync("New Password",
```
riarenas, line 50: If I understand correctly, coreidentity generates the password and then we just paste it? if so, is it worth adding this validation when we don't control the password itself?

(just something to consider if we want to not add, but I won't block this PR on this)

riarenas: Or I guess it's trying to guard from pasting some old stuff from your clipboard? hmmm

riarenas (dismissed):
Just a tiny note about the password validation

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
garath: The input gets echoed back to the console which seems to cover "old stuff on the clipboard" problem, but I could still get behind a sanity check on the input.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
```diff
@@ -30,24 +31,25 @@ protected override async Task<SecretData> RotateValue(Parameters parameters, Rot
...
+            "Expecting a valid password from CoreIdentity, containing at least 20 fairly random characters",
+            value => value != null && value.Length >= 20 && !value.All(char.IsLetterOrDigit));
 
         return new SecretData(newPassword, DateTimeOffset.MaxValue, _clock.UtcNow.AddMonths(6));
```
garath, line 54: This should be updated, too. The password expiration is enforced, and the rotation time should be a bit before that expiration (perhaps even just before the email gets sent, then the email serves as a "something's wrong" check and also doesn't cause confusion when both services are messaging a rotation).

garath (changes requested):
Requesting change to the expiration and next-rotation-on values

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
dougbu: yes, this was just intended to be a sanity check. I'm unsure whether we _should_ be echoing the input but that's a question for another week

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
dougbu: I didn't do the math to figure out when CoreIdentity decides to warn us. main thing I know is the last round of emails arrived b/4 `secret-manager` started complaining and the new expirations are about 6 months out. I'll figure the right `Add...` values b/4 merging this

dougbu:
didn't do the math on this one. just needs that single-line update

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DomainAccount.cs
garath:
```suggestion
        return new SecretData(newPassword, DateTimeOffset.MaxValue, _clock.UtcNow.AddMonths(5));
```
Just looked at the history, I think 5 months will be a good rotation time.

garath (approved):

wfurt (approved):
LGTM

davfost (approved):

Issue #5319, opened 2025-03-28
Dotneteng-status improvements - add log tables, github version running


meghnave:
Dotneteng status have been flaky for a while. Sometimes they come back up by themselves, most recently it needed a redeploy that fixed auth like issues that seemed persistent ([this issue](https://github.com/dotnet/dnceng/issues/5297)). 
So under this task, we want to cover these 2 things that came from debugging the issue and [conversation ](https://teams.microsoft.com/l/message/19:068a8eae527b4a98b71ac0c47a805a99@thread.skype/1743033287410?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=147df318-61de-4f04-8f7b-ecd328c256bb&parentMessageId=1743033287410&teamName=.NET%20Eng%20Services&channelName=V-Team%20-%20Wave%201%20General&createdTime=1743033287410) about it that would improve ops experience - 
1) Staging doesn't have AppServiceHTTPLogs and AppServiceAppLogs tables - these do exist in prod and are helpful to see errors and timings.
2) The website currently doesn't show the commit version deployed which can be helpful in [narrowing down issues ](https://teams.microsoft.com/l/message/19:068a8eae527b4a98b71ac0c47a805a99@thread.skype/1743105508030?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=147df318-61de-4f04-8f7b-ecd328c256bb&parentMessageId=1743033287410&teamName=.NET%20Eng%20Services&channelName=V-Team%20-%20Wave%201%20General&createdTime=1743105508030) as well. 
3) Add retries while getting secret - this can help reduce the crashes we see happen.

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description

Adds build number to homepage of status website

meghnave:
Task update/posterity notes:   
- PR to add buildnumber on the website- https://github.com/dotnet/dnceng/pull/5429
- Turned on the log tables that were missing in staging through the portal.
- The root cause isn't the issue with auth get, (it was a different error this time with 503s) so ended up not adding retries.
- Matched the staging slot and production slot settings for dotneteng-status-staging (which hasn't fixed the issue)
Current half-baked postulation:
[...]

PR #5429, opened 2025-04-21
Add build number on dotneteng-status homepage

meghnave:
https://github.com/dotnet/dnceng/issues/5319


meghnave:
> @dotnet-policy-service agree company="Microsoft"

meghnave:
@dotnet-policy-service agree company="Microsoft"

src/DotNet.Status.Web/Readme.Md
```diff
@@ -0,0 +1,55 @@
+(This file is entirely generated by GH copilot yay)
+# DotNet.Status.Web
+
+## Overview
+
+The `DotNet.Status.Web` service is a web application used by the .NET Engineering Services team to monitor and manage the status of engineering workflows. It provides APIs and webhooks for tracking deployments, build notifications, and other engineering processes. The service integrates with tools like Azure DevOps, GitHub, and Grafana to provide real-time status updates and annotations.
```
garath, line 6: Love this part. I think the rest could be removed as it doesn't add much useful information.

src/DotNet.Status.Web/DotNet.Status.Web/Pages/Index.cshtml
```diff
@@ -2,9 +2,11 @@
 @using System.Security.Claims
 @{
     ViewData["Title"] = "Home page";
+    var buildNumber = Environment.GetEnvironmentVariable("BuildNumber") ?? "Unknown";
```
garath, line 5: I don't think this will work. The environment at this point (runtime) is different than the environment present in deploy.yaml.

garath (changes requested):
I'm fine with using the Azure DevOps build ID as the identifier. A few more iterations and I think you'll have it.

Just for completeness, two other ideas that have been in my head:

- A version number is generated and saved during the build. You might be able to pull that. See, for example, bits like this: https://github.com/dotnet/dnceng/blob/66b8fadfc345b3fd83a0831831aef241b2dcb342/src/SecretManager/Microsoft.DncEng.SecretManager/Commands/InfoCommand.cs#L22

- Sourcelink automagically captures the build version number _and_ git commit hash. I'm not exactly sure where it gets stored (I believe it's some part of assembly info, though). Having the actual hash would be useful (but not required, as the build ID gets us there after a few steps).

src/DotNet.Status.Web/Readme.Md
garath: Also, nit, a more common filename to have the extension all lowercase.

src/DotNet.Status.Web/DotNet.Status.Web/Pages/Index.cshtml
meghnave: This is how it's looking in testing - googled about the version truncation and there's a limit of 65536 to assembly version hence that -
image

And then this from the homepage can then be tied to [this from the build logs](https://dev.azure.com/dnceng/internal/_build/results?buildId=2684186&view=logs&j=d3135522-e374-585c-cd8e-9aab495c382a&t=a4eaf635-110b-5918-2271-a3cc452aeee2&l=) - a bit long routed but could be helpful for now like we were chatting

meghnave:
> suggest checking a DLL built locally in ILSpy. its attributes should include all the interesting metadata, including a build number other than `42.42.42.42` and the `git` commit SHA
image
Ilspy seems to be showing the same too.. not sure where that's driven from.. it doesn't look hardcoded anywhere..

Also, looked at that and figured out that the assemblyinfromationalversion contains the githash so changed it to that. Could use another lookover @garath @dougbu

meghnave:
> Have you been able to run this locally and confirm the page looks proper (and doesn't crash!)?

Yup - 
image

src/DotNet.Status.Web/DotNet.Status.Web/Pages/Index.cshtml
```diff
@@ -1,10 +1,15 @@
...
     ViewData["Title"] = "Home page";
-}
+    var appVersion = Assembly.GetEntryAssembly()
+                        ?.GetCustomAttribute<AssemblyInformationalVersionAttribute>()
+                        ?.InformationalVersion?.ToString() ?? "Unknown";
+    }
 
-.NET Engineering Status Reporting endpoint
+.NET Engineering Status Reporting endpoint <br/>
+Deployed via build number: @appVersion <br/>
```
garath, line 12: This won't be a build number, right? It will be the app's version number?

meghnave: Ah yup, updated

src/DotNet.Status.Web/Readme.Md
meghnave: Updated

garath (approved):

garath (approved):
🎉

dougbu (approved):
hurrah❗

/fyi I think MSBuild (or Mr. Build) figures out the right assembly info version for itself. it'll include the commit SHA as long as nothing sets the `IncludeSourceRevisionInInformationalVersion` build property to `false`.

Issue #1430, opened 2023-11-16
Move dnceng repositories to .NET 8


garath:
The release of .NET 8 reminds us that the EOL dates for .NET 6 and 7 approach. Most of dnceng repositories are using .NET 6, but a few are using .NET 7, and that date is close enough to start thinking about allocating time to focus on an update. 

.NET 6 leaves support on November 12, 2024. .NET 7 leaves support on May 14, 2024.

| Repository                 | Framework   | Tracking or Change |
| -------------------------- | ----------- | ------------------ |
| ✅ dotnet/dnceng              | [net6.0][1] |                    |
| ✅ dotnet/dnceng-shared       | [net6.0][2] |                    |
| dotnet-helix-service       | [net7.0][3] | dotnet/dnceng#2413 | 
| ✅ dotnet-helix-machines  | [net8.0][4] | [!35018]           |
| ✅ dotnet-migrate-package | [net8.0][5] | [!39070]           |

[1]: https://github.com/dotnet/dnceng/blob/2631717c00ad0e05e3466a002f6be09d24188859/Directory.Build.props#L15C5-L15C5
[2]: https://github.com/dotnet/dnceng-shared/blob/9ee7c56ca34285d8681f750036d39b1bb7c6ac37/Directory.Build.props#L15C3-L15C3
[3]: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-service?path=/Directory.Build.props&version=GBmain&line=25&lineEnd=26&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
[4]: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-machines?path=/Directory.Build.props&version=GBmain&line=21&lineEnd=22&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
[5]: https://dev.azure.com/dnceng/internal/_git/dotnet-migrate-package?path=/migrate-package/migrate-package.csproj&version=GBmain&line=5&lineEnd=6&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents
[!39070]: https://dev.azure.com/dnceng/internal/_git/dotnet-migrate-package/pullrequest/39070
[!35018]: https://dev.azure.com/dnceng/internal/_git/dotnet-helix-machines/pullrequest/35018

### Release Note Category
- [ ] Feature changes/additions 
- [ ] Bug fixes
- [x] Internal Infrastructure Improvements
### Release Note Description
[...]

PR #5500, opened 2025-04-29
Updated to use .NET8; removed unused usings

missymessa:
https://github.com/dotnet/dnceng/issues/1430


garath:
Have you checked that any consumers of this bits published from here are also on net8?

missymessa:
@garath The repos bringing in SecretManager will have a broken PR that will need to be handled if they're not updated to .NET 8. I'm planning on handling any of those remaining, too.

garath:
So, are you planning on having those updated first so the change to secret-manager (and the other Microsoft.Dnceng.* projects, too, right?) don't break?

garath:
I see a few YAML references that need updating. 

Example, https://github.com/dotnet/dnceng/blob/66b8fadfc345b3fd83a0831831aef241b2dcb342/azure-pipelines.yml#L81

missymessa:
> I see a few YAML references that need updating.
> 
> Example,
> 
> https://github.com/dotnet/dnceng/blob/66b8fadfc345b3fd83a0831831aef241b2dcb342/azure-pipelines.yml#L81

This PR also needs an updated dnceng-shared...

garath (approved):

garath:
Heads-up, this failed in CI trying to find the monitoring project (which I believe does not get exercised in PR). [20250430.4](https://dev.azure.com/dnceng/internal/_build/results?buildId=2699496&view=logs&j=48d02e02-a4e7-5bd8-0d0c-25c42f2ba95b&t=bbe1c61a-265e-5943-8fc1-61fb5dbd1dc1&l=42)

PR #5694, opened 2025-05-20
Reformat TokenCredential to ensure it gets the pipeline's service con…

garath:
Resolves [dotnet/dnceng#5675](https://github.com/dotnet/dnceng/issues/5675)

Fix credentials used by SecretManager scenario tests (avoid letting managed identities be detected).


missymessa (approved):
LGTM shipit

PR #5775, opened 2025-05-29
Add logging to try to confirm reason for apivalidationexception ...

chcosta:
… which is being seen in the dotneteng-status app insights logs.

To see errors, look in dotneteng-status-prod appinsights logs 

> exceptions
| where problemId contains "apivalidationexception"

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
@@ -310,6 +310,13 @@ public class AzurePipelinesController : ControllerBase
                     newIssue.Labels.Add(label);
                 }
 
+                /*
+                 * We are sometimes seeing an OctoKit.ApiValidationException in the dotneteng-status app insights logs when creating issues.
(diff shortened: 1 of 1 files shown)


src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AzurePipelinesController.cs
```diff
@@ -310,6 +310,13 @@ private async Task ProcessBuildNotificationsAsync(IAzureDevOpsClient client, Bui
                     newIssue.Labels.Add(label);
                 }
 
+                /*
+                 * We are sometimes seeing an OctoKit.ApiValidationException in the dotneteng-status app insights logs when creating issues.
+                 * This is potentially related to https://github.com/octokit/octokit.net/issues/612.
+                 * Adding logging here to help us track down the issue.
+                 */
+                _logger.LogInformation("Creating issue {owner}/{repo} with title '{issueTitle}' in the {milestone} milestone.", repo.Owner, repo.Name, newIssue.Title, newIssue.Milestone);
```
garath, line 318: Could use a logging scope, which would then get attached to every following log message. But that's not a big deal, this will still make it easy to track the flow.

garath (dismissed):

garath (dismissed):

Issue #5851, opened 2025-06-10
Path forward for disabled PATs in mseng


meghnave:
[MSEng disabled PATs Apr 29th](https://eng.ms/docs/cloud-ai-platform/devdiv/one-engineering-system-1es/1es-docs/1es-security-configuration/azdo-disable-pats/mseng-pat-allowlist), it's starting to break our builds (only) now. 
[Example](https://github.com/dotnet/dnceng/issues/5842)

Filed an [exception request](https://myaccess.microsoft.com/@microsoft.onmicrosoft.com#/request-history) (check logged in as dn-bot) and reached out to Angel on chat as per the email instructions for a service bot.

We will also need to figure out the plan for moving to Entra while the exception is in place for the time being. 

### Release Note Category
- [ ] Feature changes/additions 
- [x] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
Implement workaround in secret-manager for missing account information.

missymessa:
Needs to be moved.

PR #5903, opened 2025-06-17
Add workaround for missing Azure DevOps account

garath:
Workaround for [dotnet/dnceng#5851](https://github.com/dotnet/dnceng/issues/5851)

Manually add info on the mseng org when it is requested but not found in the list of discovered accounts.


missymessa (approved):
This is sad, but... whatever works, I guess 

LGTM shipit

meghnave (approved):

Issue #6060, opened 2025-08-19
Add DotNetStatusWeb token type to Secret Manager


missymessa:
To prevent our service hook token from expiring quietly again. 

### Release Note Category
- [x] Feature changes/additions 
- [ ] Bug fixes
- [ ] Internal Infrastructure Improvements
### Release Note Description
New secret type for tracking DotNetStatusWeb tokens

missymessa:
Needs to be moved.

PR #6062, opened 2025-08-19
New secret type for DotNetStatusWeb Access Tokens

missymessa:
https://github.com/dotnet/dnceng/issues/6060

### DIFF
--- a/src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DotNetStatusWebAccessToken.cs
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DotNetStatusWebAccessToken.cs
@@ -1,0 +1,40 @@
+﻿using Microsoft.DncEng.CommandLineLib;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.DncEng.SecretManager.SecretTypes;
+
+[Name("dotnetstatusweb-access-token")]
+public class DotNetStatusWebAccessToken : GenericAccessToken
+{
+    public DotNetStatusWebAccessToken(ISystemClock clock, IConsole console) : base(clock, console)
+    {
+    }
+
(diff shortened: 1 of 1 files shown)


missymessa:
@garath I didn't write any tests for this. Didn't look like any existed for the other GenericAccessTokens types either.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DotNetStatusWebAccessToken.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public DotNetStatusWebAccessToken(ISystemClock clock, IConsole console) : base(clock, console)
+    {
+    }
+
+    private readonly string[] _expirationDateFormats = new[] { "yyyy-MM-dd", "yyyy-MM-dd HH:mm:ss" };
+
+    protected override bool HasExpiration => true;
+
+    protected override string HelpMessage => "Please login to https://{0}/Token using your GitHub account and create a new token.";
```
garath, line 20: Let's use one of the bots so the token doesn't unintentionally get deleted by a user. Say, dotnet-bot?

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DotNetStatusWebAccessToken.cs
```diff
@@ -0,0 +1,40 @@
...
+
+    protected override string TokenName => "DotNet Status Web Access Token";
+
+    protected override string TokenFormatDescription => "base64 url encoded with at least 24 characters";
```
garath, line 24: It's not a URL, it's just a big string of random bits.

src/SecretManager/Microsoft.DncEng.SecretManager/SecretTypes/DotNetStatusWebAccessToken.cs
missymessa: dotnet-bot is one we use for the secret that's kicking off the need for this, so that sounds good.

garath:
Could also have tests showing proper flow from YAML manifest through. But that's meh.

PR #6425, opened 2025-11-20
Allow major roll-forward for Secret Manager

premun:
We tried to run it with .NET 10.0 in `arcade-services` and it didn't work.

We also tried to override it and also didn't work: https://github.com/dotnet/arcade-services/pull/5546


garath:
This will break repos that aren't using net10 yet. Everyone needs to update (at least the secret-manager step) at the same time. Or maybe it would be better to multi-target? 

EDIT: this isn't quite true. See below.

garath:
Oh, yep, I had this confused with "RollForward" in global.json. [This doc](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#rollforward) says it indeed will work for runtime choices. 

I wonder if it would be better or worse to multitarget.

garath (approved):

PR #6558, opened 2026-05-19
Fix AlertHookController 401 by adding webhook secret auth

missymessa:
## Summary

The AlertHookController was returning 401 for all Grafana webhook requests because it fell under the default `msft` authorization policy requiring GitHub OAuth team membership. Grafana contact points use Basic Auth (username + password), which doesn't satisfy this policy.

This has been causing ~96 failed requests/hour since the May 13 production rollout.

## Changes

- Add `[AllowAnonymous]` to `AlertHookController` to bypass the default OAuth policy
- Add `IsAuthorized()` method that validates the Basic Auth password against a configured webhook secret from Key Vault
- Add `WebhookSecret` property to `GrafanaOptions`
- Add vault reference in `settings.json` for `grafana-webhook-secret`
- Update test to pass new `GrafanaOptions` parameter

## Deployment Steps

1. Retrieve the current Grafana contact point Basic Auth password (from Grafana API or admin UI)
2. Store it as `grafana-webhook-secret` in the dotneteng-status service's Key Vault
3. Deploy this change
4. Verify 401s stop in App Insights and alerts flow through to GitHub issues

## Related

- AzDO Work Item: https://dev.azure.com/dnceng/internal/_workitems/edit/10716
- Grafana contact points: `.NET Status Alert` (uid: ff54hqdata2gwf) and no-image variant (uid: df54hqd2wln28b)

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/.config/settings.json
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/.config/settings.json
@@ -291,7 +291,8 @@
   "Grafana": {
     "BaseUrl": "https://dotnet-eng-grafana-staging.westus2.cloudapp.azure.com",
     "ApiToken": "[vault(grafana-api-token)]",
-    "TableName": "deployments"
+    "TableName": "deployments",
+    "WebhookSecret": "[vault(grafana-webhook-secret)]"
   },
   "WebHooks": {
     "github": {
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
@@ -9,6 +10,7 @@ using System.Threading;
 using System.Threading.Tasks;
 using DotNet.Status.Web.Models;
 using DotNet.Status.Web.Options;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.DotNet.GitHub.Authentication;
 using Microsoft.Extensions.Logging;
@@ -19,6 +21,7 @@ namespace DotNet.Status.Web.Controllers;
 
 [ApiController]
 [Route("api/alert")]
+[AllowAnonymous]
 public class AlertHookController : ControllerBase
 {
     public const string NotificationIdLabel = "Grafana Alert";
@@ -32,6 +35,7 @@ public class AlertHookController : ControllerBase
 
     private readonly IOptions<GitHubConnectionOptions> _githubOptions;
     private readonly IOptions<GitHubClientOptions> _githubClientOptions;
+    private readonly IOptions<GrafanaOptions> _grafanaOptions;
     private readonly ILogger _logger;
     private readonly IGitHubTokenProvider _tokenProvider;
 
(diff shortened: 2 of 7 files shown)


src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
```diff
@@ -350,4 +362,44 @@ private async Task<IGitHubClient> GetGitHubClientAsync(string org, string repo)
...
+            string decoded = Encoding.UTF8.GetString(Convert.FromBase64String(encoded));
+            // Basic Auth format is "username:password" — we only validate the password
+            int colonIndex = decoded.IndexOf(':');
+            if (colonIndex < 0)
+            {
+                return false;
+            }
+
+            string password = decoded.Substring(colonIndex + 1);
+            return string.Equals(password, webhookSecret, StringComparison.Ordinal);
```
chcosta: i assume "webhooksecret" is not a real secret, it's just "we don't want everyone easily seeing our webhooks so we mark them as secret.  is that true?

A more secure implementation (untested) would be:

```C#
private bool IsAuthorized()
{
    var secret = _grafanaOptions.Value?.WebhookSecret;
    if (string.IsNullOrEmpty(secret))
    {
        _logger.LogError("Grafana WebhookSecret is not configured");
        return false;
    }

    if (!Request.Headers.TryGetValue("Authorization", out var authHeader))
        return false;

    if (!authHeader.ToString().StartsWith("Basic ", StringComparison.OrdinalIgnoreCase))
        return false;

    try
    {
        var encoded = authHeader.ToString().Substring("Basic ".Length).Trim();
        var bytes = Convert.FromBase64String(encoded);

        try
        {
            int colon = Array.IndexOf(bytes, (byte)':');
            if (colon < 0) return false;

            var provided = bytes.AsSpan(colon + 1);
            var expected = Encoding.UTF8.GetBytes(secret);

            return CryptographicOperations.FixedTimeEquals(provided, expected);
        }
        finally
        {
            Array.Clear(bytes, 0, bytes.Length);
        }
    }
    catch
    {
        return false;
    }
}
```

that is slightly more secure because it prevents against timing attacks and unnecessary string exposure.  I think the recommended approach is Grafana HMAC verification.

src/DotNet.Status.Web/DotNet.Status.Web/.config/settings.json
```diff
@@ -291,7 +291,8 @@
   "Grafana": {
     "BaseUrl": "https://dotnet-eng-grafana-staging.westus2.cloudapp.azure.com",
     "ApiToken": "[vault(grafana-api-token)]",
-    "TableName": "deployments"
+    "TableName": "deployments",
+    "WebhookSecret": "[vault(grafana-webhook-secret)]"
```
haruna99, line 295: This secret does not exist in the manifest which is causing the PR to fail

missymessa: Fixed — added `grafana-webhook-secret` as a `text` type secret to both `dotneteng-status-prod.yaml` and `dotneteng-status-staging.yaml` vault configs.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/AlertHookController.cs
missymessa: Good call — updated to use `CryptographicOperations.FixedTimeEquals` with raw byte comparison and `Array.Clear` after use. Avoids both timing attacks and unnecessary string materialization of the secret.

missymessa:
## Why not just fix the existing OAuth policy?

The `msft` authorization policy (applied globally via `DefaultAuthorizeActionModelConvention` in Startup.cs) requires GitHub OAuth team membership — it's designed for browser-based human users accessing the Status Web UI.

Grafana contact points are server-to-server webhooks that POST with Basic Auth credentials. There's no way to make Grafana authenticate via GitHub OAuth — it doesn't have a browser session or GitHub team membership.

**Timeline:**
- **Feb 2026** — Stu added AlertHookController + Grafana alert integration (commit cc61fa12). The controller inherited the default `msft` policy but wasn't yet exercised in production.
- **Apr–May 2026** — Managed Identity migration PRs (#6480, #6490, #6524) triggered deployments that included the new controller.
- **May 13** — Production deployment activated the endpoint; Grafana started hitting it and getting 401s (~96/hour) because Basic Auth can't satisfy GitHub OAuth.

**Conclusion:** The `msft` policy was never compatible with webhook delivery. `[AllowAnonymous]` + explicit Basic Auth secret validation is the correct pattern for machine-to-machine webhook endpoints (same pattern used by the existing `WebHookController` for GitHub webhooks).

missymessa:
@chcosta @haruna99 Both feedback items addressed in the latest push — timing-safe comparison + vault manifest entries added. Could one of you take another look when you get a chance? This is blocking the Grafana alert (firing 5+ days now). Thanks!

haruna99 (approved):

PR #6597, opened 2026-05-29
One pager for epic 10788

davfost:
Replace image creation logic with a standalone Image Creation Subsystem:
https://dnceng.visualstudio.com/internal/_workitems/edit/10788


davfost:
Abandoning and moving one pager to the helix machines repo

PR #6627, opened 2026-07-02
Remove service-connections-devdiv.yaml (WI 10115 cleanup)

missymessa:
## Summary

Deletes `.vault-config/service-connections-devdiv.yaml` — the secret-manager manifest for the `dnceng-test-tools-feed` service connection in devdiv/DevDiv.

## Why

- The `dnceng-test-tools-feed` SC is **packaging-only and exempt** from the PAT disable policy
- WIF migration was attempted but failed (`ExternalNuGetFeed` type is incompatible with `workloadidentityuser` SCs)
- The Entra app registration and service principal created during the attempt have already been deleted
- This manifest entry serves no purpose — the SC is not managed by secret-manager rotation (it's a PAT-based ExternalNuGetFeed endpoint in DevDiv org)

## Work Item

Closes cleanup for [WI 10115](https://dnceng.visualstudio.com/internal/_workitems/edit/10115)


missymessa:
Closing — this was premature. The SC still needs to be migrated, not abandoned. Will re-approach with an AzureRM WIF SC (the prior attempt used the wrong SC type).

PR #6643, opened 2026-07-17
Make dotnet-eng-status deployment notifications best-effort (fix intermittent 500)

missymessa:
## Problem

The Helix deploy pipelines (`dotnet-helix-service-deploy`, `dotnet-helix-machines` deploy) intermittently fail their **"Notify dotnet-eng-status/start"** step with an HTTP 500, e.g. [build 3024453](https://dnceng.visualstudio.com/internal/_build/results?buildId=3024453&view=results):

```
POST https://dotneteng-status-staging.azurewebsites.net/api/deployment/dotnet-helix-service/2026071601/start -> (500) Internal Server Error
```

The 500 comes from `DeploymentController.MarkStart` in the DotNetStatus (dotnet-eng-status) app, **not** from Helix. Recording a deployment annotation is best-effort telemetry, but the controller had two problems that let any hiccup become a 500 that shows up as a red step in the deploy pipeline:

1. The Grafana annotation call was wrapped in `ExponentialRetry.RetryAsync(..., isRetryable: e => true)`, which retries **every** exception (including non-transient 4xx from `EnsureSuccessStatusCode()`) up to 10 times and then rethrows.
2. Neither the retried Grafana call nor the follow-up table upsert was wrapped in a try/catch, so any rethrown/unretried exception became an unhandled 500.

## Fix

- Narrow the retry predicate to a new `IsTransientFailure` helper: retry only network errors, timeouts, and 5xx responses (`HttpRequestException` with `StatusCode` null or >= 500, or `TaskCanceled`/`OperationCanceled`). Non-transient failures now fail fast instead of being retried 10× and rethrown.
- Wrap the `MarkStart` and `MarkEnd` bodies in try/catch that logs the error and returns `NoContent()`. Best-effort deployment notifications never fail the calling pipeline again.
- `MarkEnd` now logs and returns `NoContent()` instead of `NotFound()` when no matching start record exists (a missing/failed start notification should not turn the end notification into a 404).

## Notes

- Behavior on the success path is unchanged.
- Observability on the staging app is currently broken (App Insights instrumentation key points at an empty component), so the exact failing dependency (Grafana POST vs. table upsert) could not be captured from telemetry; this fix hardens both paths regardless.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
+++ b/src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
@@ -4,26 +4,15 @@
 
 using Azure;
 using System;
-using System.Collections.Immutable;
 using System.ComponentModel.DataAnnotations;
-using System.IO;
-using System.Net;
-using System.Net.Http;
-using System.Net.Http.Headers;
-using System.Text;
-using System.Threading;
 using System.Threading.Tasks;
 using Azure.Data.Tables;
 using DotNet.Status.Web.Options;
 using Microsoft.AspNetCore.Mvc;
-using Microsoft.DotNet.Services.Utility;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
-using Newtonsoft.Json;
-using Newtonsoft.Json.Serialization;
 using StatusWebAnnotationEntity = DotNet.Status.Web.Models.AnnotationEntity;
-using Microsoft.WindowsAzure.Storage.Table;
 using Azure.Identity;
 
 namespace DotNet.Status.Web.Controllers;
@@ -33,17 +22,14 @@ namespace DotNet.Status.Web.Controllers;
 public class DeploymentController : ControllerBase
 {
     private readonly IHostEnvironment _env;
-    private readonly ExponentialRetry _retry;
     private readonly IOptionsMonitor<GrafanaOptions> _grafanaOptions;
     private readonly ILogger<DeploymentController> _logger;
 
     public DeploymentController(
-        ExponentialRetry retry,
         IOptionsMonitor<GrafanaOptions> grafanaOptions,
         ILogger<DeploymentController> logger,
         IHostEnvironment env)
     {
-        _retry = retry;
         _grafanaOptions = grafanaOptions;
         _logger = logger;
         _env = env;
@@ -53,57 +39,24 @@ public class DeploymentController : ControllerBase
     public async Task<IActionResult> MarkStart([Required] string service, [Required] string id)
     {
         _logger.LogInformation("Recording start of deployment of '{service}' with id '{id}'", service, id);
-        NewGrafanaAnnotationRequest content = new NewGrafanaAnnotationRequest
-        {
-            Text = $"Deployment of {service}",
-            Tags = new[] {"deploy", $"deploy-{service}", service},
-            Time = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds(),
-        };
 
-        // Recording a deployment annotation is best-effort telemetry and must never fail the
-        // calling deployment pipeline. Any failure talking to Grafana or table storage is logged
-        // and swallowed so the notification endpoint returns success instead of a 500.
+        // Recording a deployment annotation is best-effort telemetry and must never fail the calling
+        // deployment pipeline. The annotation is written to Azure Table storage, which Grafana reads
+        // directly; any failure is logged and swallowed so the endpoint returns success instead of a 500.
         try
         {
-            NewGrafanaAnnotationResponse annotation;
-            using (var client = new HttpClient(new HttpClientHandler() { CheckCertificateRevocationList = true }))
-            {
-                annotation = await _retry.RetryAsync(async () =>
-                    {
-                        GrafanaOptions grafanaOptions = _grafanaOptions.CurrentValue;
-                        _logger.LogInformation("Creating annotation to {url}", grafanaOptions.BaseUrl);
-                        using (var request = new HttpRequestMessage(HttpMethod.Post,
-                                   $"{grafanaOptions.BaseUrl}/api/annotations"))
-                        {
-                            request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
-                            request.Headers.Authorization =
-                                new AuthenticationHeaderValue("Bearer", grafanaOptions.ApiToken);
-                            request.Content = CreateObjectContent(content);
-
-                            using (HttpResponseMessage response = await client.SendAsync(request, CancellationToken.None))
-                            {
-                                _logger.LogTrace("Response from grafana {responseCode} {reason}", response.StatusCode, response.ReasonPhrase);
-                                response.EnsureSuccessStatusCode();
-                                return await ReadObjectContent<NewGrafanaAnnotationResponse>(response.Content);
-                            }
-                        }
-                    },
-                    e => _logger.LogWarning(e, "Failed to send new annotation"),
-                    IsTransientFailure
-                );
-            }
-            _logger.LogInformation("Created annotation {annotationId}, inserting into table", annotation.Id);
-
             TableClient table = await GetCloudTable();
-            await table.UpsertEntityAsync(new DotNet.Status.Web.Models.AnnotationEntity(service, id, annotation.Id)
+            await table.UpsertEntityAsync(new StatusWebAnnotationEntity(service, id)
             {
+                Started = DateTimeOffset.UtcNow,
                 ETag = new Azure.ETag("*")
             });
         }
         catch (Exception ex)
         {
-            _logger.LogError(ex, "Failed to record start of deployment of '{service}' with id '{id}'; continuing without blocking the deployment", service, id);
+            _logger.LogError(ex, "Failed to record start of deployment of '{service}' with id '{id}'", service, id);
         }
+
         return NoContent();
     }
 
@@ -111,8 +64,10 @@ public class DeploymentController : ControllerBase
     public async Task<IActionResult> MarkEnd([Required] string service, [Required] string id)
     {
         _logger.LogInformation("Recording end of deployment of '{service}' with id '{id}'", service, id);
-        // As with MarkStart, recording the end of a deployment is best-effort telemetry and must
-        // never fail the calling deployment pipeline. Any failure is logged and swallowed.
+
+        // As with MarkStart, recording the end of a deployment is best-effort telemetry written to
+        // Azure Table storage and must never fail the calling deployment pipeline. Any failure is
+        // logged and swallowed.
         try
         {
             TableClient table = await GetCloudTable();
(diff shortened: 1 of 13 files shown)


chcosta:
> Recording a deployment annotation 

what's the deployment annotation used for?  where does it surface?  

> Narrow the retry predicate to a new IsTransientFailure helper: retry only network errors, timeouts, and 5xx responses (HttpRequestException with StatusCode null or >= 500, or TaskCanceled/OperationCanceled). Non-transient failures now fail fast instead of being retried 10× and rethrown.

this seems good

> Wrap the MarkStart and MarkEnd bodies in try/catch that logs the error and returns NoContent(). Best-effort deployment notifications never fail the calling pipeline again.

isn't this just masking the failure?  if we're just going to do this, then it sounds like the notifications aren't needed.  can we just remove them?  was this for rollout scorecard or something else?

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
```diff
@@ -58,98 +59,131 @@ public async Task<IActionResult> MarkStart([Required] string service, [Required]
...
-                e => true
-            );
+                    },
+                    e => _logger.LogWarning(e, "Failed to send new annotation"),
+                    IsTransientFailure
+                );
+            }
+            _logger.LogInformation("Created annotation {annotationId}, inserting into table", annotation.Id);
+
+            TableClient table = await GetCloudTable();
```
garath, line 48: Consider deleting the direct write to the Grafana service and just keep the write to Azure Tables. Grafana can read the information from there.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
```diff
@@ -58,98 +59,131 @@ public async Task<IActionResult> MarkStart([Required] string service, [Required]
...
         }
-        _logger.LogInformation("Created annotation {annotationId}, inserting into table", annotation.Id);
-
-        TableClient table = await GetCloudTable();
-        await table.UpsertEntityAsync(new DotNet.Status.Web.Models.AnnotationEntity(service, id, annotation.Id)
+        catch (Exception ex)
         {
-            ETag = new Azure.ETag("*")
-        });
+            _logger.LogError(ex, "Failed to record start of deployment of '{service}' with id '{id}'; continuing without blocking the deployment", service, id);
```
garath:
```suggestion
            _logger.LogError(ex, "Failed to record start of deployment of '{service}' with id '{id}'", service, id);
```
No need to overload the message

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
```diff
@@ -58,98 +59,131 @@ public async Task<IActionResult> MarkStart([Required] string service, [Required]
...
+
+            var updateResult = await table.UpdateEntityAsync(annotation, annotation.ETag, TableUpdateMode.Replace);
+            _logger.LogInformation("Update response code {responseCode}", updateResult.Status);
+
+            var content = new NewGrafanaAnnotationRequest
+            {
+                TimeEnd = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds(),
+            };
+
+            using (var client = new HttpClient(new HttpClientHandler() { CheckCertificateRevocationList = true }))
```
garath: Same as above. I'm okay deleting this whole direct-write-to-Grafana section and relying only on the Azure Table store.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
```diff
@@ -58,98 +59,131 @@ public async Task<IActionResult> MarkStart([Required] string service, [Required]
...
-            );
+                    },
+                    e => _logger.LogWarning(e, "Failed to send new annotation"),
+                    IsTransientFailure
+                );
+            }
+        }
+        catch (Exception ex)
+        {
+            _logger.LogError(ex, "Failed to record end of deployment of '{service}' with id '{id}'; continuing without blocking the deployment", service, id);
```
garath:
```suggestion
            _logger.LogError(ex, "Failed to record end of deployment of '{service}' with id '{id}'", service, id);
```

epananth:
There was a transient error where one will have to go delete a file in Azure portal for this service to work. I think Haruna and Meghna did that a few times.

It would be nice if this could cover that case too

garath (changes requested):

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
missymessa: Applied the suggestion — simplified the log message to: LogError(ex, "Failed to record start of deployment of '{service}' with id '{id}'", service, id).

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
missymessa: Applied the suggestion — simplified the log message to: LogError(ex, "Failed to record end of deployment of '{service}' with id '{id}'", service, id).

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
missymessa: Done — removed the direct write to the Grafana annotations API from MarkStart; it now only writes the annotation to the Azure Table (which Grafana reads from). This also drops the Grafana ApiToken/retry HTTP path that was the source of the intermittent 500s.

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
missymessa: Done — deleted the entire direct-write-to-Grafana section in MarkEnd too; it now only updates the Ended time on the Azure Table record. Also removed the now-unused IsTransientFailure, retry, HttpClient, and Grafana request/response DTOs.

missymessa:
Good questions — and the review consensus (thanks @garath) actually resolves the masking concern by changing the approach:

**What the annotation is / where it surfaces:** the deployment annotation is a Grafana dashboard annotation — a vertical marker on the time-series dashboards showing when each service was deployed (start→end span, tagged `deploy`, `deploy-<service>`, `<service>`). It's used to correlate metric/behavior changes with deployments while looking at Grafana (not a rollout scorecard).

**On "isn't this just masking the failure?":** you're right that swallowing a failing Grafana push would just hide a broken notification. So instead of masking it, this revision **removes the fragile direct call to the Grafana annotations API entirely**. `MarkStart`/`MarkEnd` now only write the deployment record to the **Azure Table**, and Grafana reads the annotations from that table directly. That deletes the HTTP dependency (and its `ApiToken` + 10× retry-then-rethrow path) that was actually producing the intermittent 500s.

So we keep the useful signal (the durable table record that Grafana renders), and the endpoint no longer has a flaky external call to mask. The best-effort try/catch now only guards the Azure Table write (managed-identity, SDK-retried), so a genuine storage outage still won't fail the deploy pipeline, but there's no longer a Grafana-push failure being hidden. Pushed in b4f7c83d.

missymessa:
Thanks @epananth — that "have to go delete a file in the Azure portal to get the service working again" case is the leftover **`app_offline.htm`**, and it's already been fixed separately in #6580 (merged 2026-05-22, WI 8535).

Root cause there: the `AzureRmWebAppDeployment@4` task defaults `TakeAppOfflineFlag` to `true`, which drops an `app_offline.htm` during deploy; with zipDeploy that's unnecessary (the deploy is atomic) and intermittently the file wasn't cleaned up, leaving the app offline until someone deleted it manually (what Haruna/Meghna were doing). #6580 sets `TakeAppOfflineFlag: false` in `eng/deploy.yaml` so the file is never created.

So that failure mode is covered by #6580; this PR (#6643) is the separate fix for the intermittent **500s** from the deployment-notification endpoint. Let me know if you've seen the app_offline.htm situation recur *after* #6580 shipped — if so I'll reopen that thread.

missymessa:
> There was a transient error where one will have to go delete a file in Azure portal for this service to work. I think Haruna and Meghna did that a few times.
> 
> It would be nice if this could cover that case too

@epananth do you still see that happening? I put in a fix for that a while ago: https://github.com/dotnet/dnceng/pull/6580

src/DotNet.Status.Web/DotNet.Status.Web/Controllers/DeploymentController.cs
```diff
@@ -32,17 +22,14 @@ namespace DotNet.Status.Web.Controllers;
 public class DeploymentController : ControllerBase
 {
     private readonly IHostEnvironment _env;
-    private readonly ExponentialRetry _retry;
     private readonly IOptionsMonitor<GrafanaOptions> _grafanaOptions;
```
garath: Still needed even though communication with Grafana has been removed?

missymessa: Yes, still needed — but only for the Azure Table coordinates, not for talking to Grafana. `_grafanaOptions` is read in `GetCloudTable()` for `options.TableName` and `options.TableUri` to construct the `TableClient`, which is where `MarkStart`/`MarkEnd` now write the deployment record. That's the same table `AnnotationsController` reads back and serves to Grafana (`/api/annotations/**`), so the field is genuinely still used.

What the Grafana removal made unused is only the HTTP-specific fields on `GrafanaOptions` (`BaseUrl`/`ApiToken`) *within this controller* — but those fields live on the shared `GrafanaOptions` type that `AnnotationsController` and `AlertHookController` also bind, so I left `GrafanaOptions` intact and can't drop the field here (removing it would break `GetCloudTable`).

Agreed the name is now a bit of a misnomer (it carries table config, not just Grafana-API config). Happy to rename/split `GrafanaOptions` into a dedicated `DeploymentTableOptions` (or similar) as a small follow-up, but that touches `AnnotationsController`/`AlertHookController`/`Startup`/tests, so I kept it out of this focused 500-fix PR. Let me know if you'd prefer I fold that cleanup in here.

garath: Yes, probably best to rename so that the file is consistent.

missymessa: Done in 3560e7e9. Split the deployment table config out of `GrafanaOptions` into a dedicated `DeploymentTableOptions` (`TableUri`/`TableName`), bound to a new `"Deployments"` config section — matching the existing `AzureTableTokenStoreOptions` pattern. Both `DeploymentController` and `AnnotationsController` (which read/write the same deployments table) now use it, so neither references a "Grafana"-named type anymore.

`GrafanaOptions` is now just `WebhookSecret` (still used by `AlertHookController`); I also removed the now-dead `BaseUrl`/`ApiToken` fields and their `settings.json` entries, since those only fed the removed Grafana annotations HTTP calls. Build + controller tests pass.

garath (approved):

garath (changes requested):
Oh, sorry for swapping the review status. I just noticed that removing `grafana-api-token` in the config also removes its need to be in the secret-manager manifest. That's a manually-managed secret so removing it is a big win for overhead.

garath:
> Oh, sorry for swapping the review status. I just noticed that removing `grafana-api-token` in the config also removes its need to be in the secret-manager manifest. That's a manually-managed secret so removing it is a big win for overhead.

These secrets are no longer needed:

- https://github.com/dotnet/dnceng/blob/3f001e477d1e91b013cad5b19f9f2d127e5b93a8/.vault-config/dotneteng-status-local.yaml#L22-L25
- https://github.com/dotnet/dnceng/blob/3f001e477d1e91b013cad5b19f9f2d127e5b93a8/.vault-config/dotneteng-status-prod.yaml#L22-L26
- https://github.com/dotnet/dnceng/blob/3f001e477d1e91b013cad5b19f9f2d127e5b93a8/.vault-config/dotneteng-status-staging.yaml#L22-L26

missymessa:
Done in 189ac85 — removed the `grafana-api-token` secret from all three vault-config manifests:

- `.vault-config/dotneteng-status-local.yaml`
- `.vault-config/dotneteng-status-prod.yaml`
- `.vault-config/dotneteng-status-staging.yaml`

Confirmed it's safe: that secret was only consumed by the DotNet.Status.Web Grafana-annotations HTTP call (now removed). The dashboard-deploy pipeline (`eng/deploy-managed-grafana.yml`) uses a separate `grafana-admin-api-key` in the AMG key vault, and `grafana-webhook-secret` (AlertHookController) is untouched. Good call on dropping the manually-managed secret.

garath (approved):

PR #6644, opened 2026-07-17
Prompt for GitHub access token expiration duration during rotation

garath:
## Summary

The `github-access-token` SecretManager secret type previously hard-coded a 90-day expiration (display only) and a fixed 30-day rotation, with `ExpiresOn` set to `DateTimeOffset.MaxValue`. Operators now enter the token's expiration **duration in days** interactively during rotation, and that value drives the on-screen GitHub instructions, the secret's real `ExpiresOn`, and its `NextRotationOn`.

## Changes

- **`SecretTypes/GitHubAccessToken.cs`**
  - Prompts for expiration duration via the existing `PromptAndValidateAsync<int>` helper.
  - Validates a whole number of days in **7–30** (we deliberately restrict below GitHub's allowed maximum); out-of-range/invalid input prints an error and re-prompts (3 attempts, then throws).
  - `ExpiresOn = now + duration`; `NextRotationOn = now + duration * 2/3` (rotate when ~1/3 of the lifetime remains).
  - On-screen GitHub guidance and a "next rotation" line reflect the entered duration.
  - Non-interactive runs still throw `HumanInterventionRequiredException`; PAT length check and login flow unchanged.
  - Uses explicit types (no `var`).
- **`Readme.md`** — documents the new prompt and the 7–30 day range.
- **New `Microsoft.DncEng.SecretManager.Tests/GitHubAccessTokenTests.cs`** — 17 tests covering duration-bound validation (min/max/negative/non-numeric/non-integer) and the rotation/expiration date math.

## Validation

`dotnet test` on `Microsoft.DncEng.SecretManager.Tests` (filtered to `GitHubAccessTokenTests`): **17 passed, 0 failed**.

### DIFF
--- a/src/SecretManager/Microsoft.DncEng.SecretManager.Tests/GitHubAccessTokenTests.cs
+++ b/src/SecretManager/Microsoft.DncEng.SecretManager.Tests/GitHubAccessTokenTests.cs
@@ -19,10 +19,12 @@ public class GitHubAccessTokenTests
     }
 
     [Test]
-    [TestCase("1", true, 1, Description = "Minimum allowed duration")]
+    [TestCase("7", true, 7, Description = "Minimum allowed duration")]
     [TestCase("90", true, 90, Description = "Typical duration")]
     [TestCase("366", true, 366, Description = "Maximum allowed duration")]
-    [TestCase("0", false, 0, Description = "Below minimum")]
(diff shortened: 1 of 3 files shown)


missymessa (approved):
LGTM shipit

PR #6650, opened 2026-07-22
Add accessible data-table equivalents for Grafana home dashboard graphs (AzDO WI 10233)

missymessa:
Addresses accessibility bug [AzDO #10233](https://dev.azure.com/dnceng/internal/_workitems/edit/10233) (MAS 1.3.1 - Info and Relationships): screen readers (Narrator/NVDA) announce nothing for the Grafana home dashboard graph panels because timeseries panels render into an opaque HTML canvas.

## Why a prior attempt was insufficient
An earlier fix only added a Grafana panel `description` field. In Grafana the description renders as an info-icon tooltip in the panel header - it does not become the visualization's accessible name and is not announced while navigating the graph. The canvas remained opaque. (Validated with the a11y agent against MAS 1.3.1.)

## Changes
`src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json`

- Added two screen-reader navigable **Table panels** presenting the same data as the graphs, using the identical Kusto queries with `resultFormat=table`:
  - "Work Items Waiting Time (Build Pools) - Accessible Data Table" (id 12)
  - "Work Items Waiting Time (Test Queues) - Accessible Data Table" (id 13)
- Added programmatically-meaningful `description` fields to the two graph panels (id 4, id 10) summarizing purpose, series, units, thresholds, and pointing to the accessible tables.

Grafana Table panels render a real HTML `<table>` with column headers, making the 95th/50th percentile waiting-time data programmatically determinable and navigable by assistive tech - satisfying MAS 1.3.1.

## Testing
- JSON validated (parses cleanly, 9 panels).
- Recommend re-testing with Narrator and NVDA in both scan modes to confirm the table headers and cell values are announced.

### DIFF
--- a/src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
+++ b/src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
@@ -284,6 +284,7 @@
           "visible": true
         }
       ],
+      "description": "Time series graph of Build Pool work item waiting times. Two series are plotted in minutes per day: Percentile95 (95th percentile) and Percentile50 (median). A red threshold line marks 30 minutes, above which waiting times are considered too high. An accessible, screen-reader navigable data table with the same values is provided in the \"Work Items Waiting Time (Build Pools) - Accessible Data Table\" panel below (MAS 1.3.1 - Info and Relationships).",
       "title": "Work Items Waiting Time (Build Pools)",
       "type": "timeseries"
     },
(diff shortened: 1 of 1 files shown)


epananth:
Could you add screenshots showing how the updated dashboard and accessible table alternatives appear in Grafana? If possible, please include both the default view and the table after scrolling through the full four-month range.

missymessa:
Thanks @epananth! Here are screenshots of the updated dashboard and the accessible data-table alternatives, rendered over the full four-month range (the dashboard's default **Last 4 months** window).

**Default dashboard view** — each time-series graph now has an accessible data-table equivalent directly beneath it:

Default dashboard view (Last 4 months)

**Original graph — Work Items Waiting Time (Test Queues):**

Test Queues graph

**Accessible data-table alternative (Test Queues)** — the same data as a table, scrolled through the full four-month range, with the percentile selector:

Test Queues accessible data table

_Note: the **Build Pools** graph and its accessible table both show "No data" for this window — they share an identical query (`QueueName contains "buildpool"`) that currently has no matching rows, so the table faithfully mirrors the graph._

garath:
Consider just deleting the graphs on the home dashboard. I don't think they're used by anyone. They reflected a metric from before the Snap.

garath (approved):

PR #6663, opened 2026-07-30
Add alerting for stalled TimelineBuilds data ingestion (AzDO 10698)

missymessa:
## What
Adds a Grafana **data-freshness alert** — `TimelineBuilds Data Ingestion Stalled` — in both **Production** and **Staging**, captured in code under `src/Monitoring/Monitoring.DncEng/alertrules/` so it deploys through the normal monitoring pipeline (not hand-applied in Grafana).

Fixes the monitoring gap called out in **AzDO WI 10698**.

## Why
The `Microsoft.DotNet.AzureDevOpsTimeline` ingestion pipeline (which fills the `engineeringdata.TimelineBuilds` Kusto table) stalled silently on **2026-04-01** and went unnoticed for **weeks** because nothing monitored data freshness. The underlying outage was already fixed (dotnet/dnceng#6540/#6541); this PR adds the missing alerting so a future stall is caught in ~2 hours instead of weeks.

## How it works
- Query uses Kusto `ingestion_time()` (the table has no dedicated ingestion column) to compute minutes since the last ingested row.
- Fires when `AgeMinutes > 120`. Normal cadence is ~1 batch/hour (verified: 7-day P95 gap 61 min, **max gap 70 min**), so 120 min gives headroom without false positives.
- `noDataState = Alerting` so a **total** stall (no rows in the 6h window) also fires.
- Routed to the **`Teams Alert`** contact point (FR/Ops Teams channel) for actionable on-call notification.

## Verification (live, both clusters)
| Cluster (env) | Last ingest | Rows/24h |
|---|---|---|
| `engsrvprod` (Production) | ~5 min ago | 2,152 |
| `engdata` (Staging) | ~16 min ago | 2,120 |

Both ADX datasources (`OlcfOPi7z`) point at live TimelineBuilds tables, so both env alerts are valid and won't false-fire. Embedded query validated end-to-end (returns a single numeric `AgeMinutes`). JSON validated; mirrors the existing Kusto alertrule schema (e.g. `helix-autoscaler-service-stopped`, `work-items-waiting-time-test-queues`).

## Notes
- No path filter / service code changes — pure monitoring config.
- Threshold (120 min) and receiver (`Teams Alert`) are easy to tune if desired.

Closes AzDO WI 10698.

### DIFF
--- a/src/Monitoring/Monitoring.DncEng/alertrules/Production/timeline-builds-ingestion-stalled.alert.json
+++ b/src/Monitoring/Monitoring.DncEng/alertrules/Production/timeline-builds-ingestion-stalled.alert.json
@@ -1,0 +1,113 @@
+{
+  "uid": "timeline-builds-ingestion-stalled",
+  "title": "TimelineBuilds Data Ingestion Stalled",
+  "condition": "C",
+  "data": [
+    {
+      "refId": "A",
+      "queryType": "",
+      "datasourceUid": "OlcfOPi7z",
+      "relativeTimeRange": {
+        "from": 21600,
+        "to": 0
+      },
+      "model": {
+        "database": "engineeringdata",
+        "datasource": {
+          "type": "grafana-azure-data-explorer-datasource",
+          "uid": "OlcfOPi7z"
+        },
+        "expression": {
+          "groupBy": {
+            "expressions": [],
+            "type": "and"
+          },
+          "reduce": {
+            "expressions": [],
+            "type": "and"
+          },
+          "where": {
+            "expressions": [],
+            "type": "and"
+          }
+        },
+        "query": "TimelineBuilds\n| where ingestion_time() > ago(6h)\n| summarize LastIngest = max(ingestion_time())\n| extend AgeMinutes = todouble(datetime_diff('minute', now(), LastIngest))\n| project AgeMinutes",
+        "querySource": "raw",
+        "rawMode": true,
+        "refId": "A",
+        "resultFormat": "table"
+      }
+    },
+    {
+      "refId": "B",
+      "queryType": "",
+      "datasourceUid": "-100",
+      "relativeTimeRange": {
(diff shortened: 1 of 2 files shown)


mmitche:
@missymessa Do you happen to know what cases we use this data for, and how long we need data for? I think we may be able switch this to use CloudMine's data.

mmitche (approved):

missymessa:
@mmitche 

What produces it: The  AzureDevOpsTimeline  Service Fabric service ( src/Telemetry/AzureDevOpsTimeline ) ingests AzDO build + timeline data for our orgs into three tables —  TimelineBuilds ,  TimelineRecords ,  TimelineIssues . It's not a raw mirror; it adds dnceng-specific enrichment that raw AzDO/CloudMine data won't have:
•  ImageName  scraped from build logs ( BuildLogScraper )
• normalized branch names
• issue bucketing/categorization (the known-build-error grouping)
• build-request validation messages
What consumes it:
• DotNet.Status.Web build-failure triage ( TimelineIssueTriage.cs ) — creates/updates the known-build-error GitHub issues and writes the  TimelineIssuesTriage  table, which feeds our PowerBI Build Analysis reports. This is the main product-facing consumer.
• The ingestion service's own checkpoint ( GetLatestTimelineBuild  =  max(FinishTime) ) to know where to resume.
• Ad-hoc Kusto / PowerBI build-analysis queries.
• Notably no Grafana dashboard queries these tables (grep confirms only this new alert references them).
How long we keep it: Retention is 200 days soft-delete (recoverability disabled). Table currently spans ~2026‑01‑05 → now (~362K rows). The PowerBI trend reporting is the longest-lookback consumer; I don't have a hard number on how far back those reports actually reach, but 200 days has been the effective window.

--------

I can open an issue for us to look into moving this to cloudmine eventually, just need to make sure we are able to get all the data we need for the alert

PR #6679, opened 2026-08-11
Retire unused Grafana home dashboard (WI 10233)

missymessa:
Addresses [AzDO work item 10233](https://dev.azure.com/dnceng/internal/_workitems/edit/10233).

The Grafana `home` dashboard is unused, so this PR now retires it instead of continuing to remediate its inaccessible graph panels.

## Changes

- Delete the managed `home` dashboard definition. The monitoring publisher removes managed dashboards that are no longer present during deployment.
- Clear the Grafana organization home-dashboard preference so it falls back to the default homepage instead of retaining the deleted UID.
- Preserve the production and staging work-item waiting-time alert rules.
- Remove those alert rules' dashboard and panel associations so they do not link to deleted panels.

## Validation

- All modified JSON files parse successfully.
- No alert rule still references dashboard UID `home`.
- `dotnet test src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/Microsoft.DotNet.Monitoring.Sdk.Tests.csproj`: 8 passed.

## Production verification

After deployment, verify that dashboard UID `home` no longer exists, Grafana opens its default homepage, and both work-item waiting-time alert rules remain enabled.


src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
```diff
@@ -246,7 +189,7 @@
           "querySource": "raw",
           "rawMode": true,
           "refId": "A",
-          "resultFormat": "time_series"
+          "resultFormat": "table"
```
garath: This is suspect. Changing a time_series to table seems like it will break things.

src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
```diff
@@ -491,9 +377,9 @@
           "visible": true
         }
       ],
-      "description": "Time series graph of Test Queue work item waiting times. Two series are plotted in minutes per day: Percentile95 (95th percentile) and Percentile50 (median). A red threshold line marks 35 minutes, above which waiting times are considered too high. An accessible, screen-reader navigable data table with the same values is provided in the \"Work Items Waiting Time (Test Queues) - Accessible Data Table\" panel below (MAS 1.3.1 - Info and Relationships).",
+      "description": "Screen-reader navigable table of Test Queue work item waiting times in minutes per day. It presents the 95th percentile (Percentile95) and median (Percentile50). Values above the 35 minute threshold are considered too high.",
```
garath: Time series is very different than a table. Won't a table lose the "time" element of it? We won't be able to view the data over time?

garath (changes requested):

src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
missymessa: Addressed in 054de0d0. The accessible tables retain the `Queued` date column for every observation, and the lower panels are now explicitly labeled Visual Trend time-series panels showing the same Percentile95/Percentile50 data over time. This preserves both screen-reader access at the tested focus point and the visual trend analysis.

src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
missymessa: Addressed in 054de0d0. The table-format targets remain only on the keyboard-focused semantic table panels (IDs 4 and 10). Panels 12 and 13 now preserve the same queries as time-series visualizations with `resultFormat: time_series`, so the trend view is not lost. The production alerts are independently deployed unified alert rules and remain linked to IDs 4 and 10.

src/Monitoring/Monitoring.DncEng/dashboard/general/home.dashboard.json
garath: This is hard for me to visualize. Can you share a screen cap of the before and after? That would help a lot.

garath:
Ah, nevermind on my reviews so far. I see what's happening. There are _new_ tables to supplement the graphs. I'm not sure how useful a table with hundreds of rows is for a screen-reader but I suppose it matches the letter of the law. 

I suggest a pivot: This dashboard isn't used. Let's just delete it.

missymessa:
Pivoted per review feedback: the unused dashboard is now deleted, the organization home preference is cleared, and the independent queue waiting-time alerts remain active without stale dashboard links.

MichaelSimons:
TIL and was surprised an accessibility compliance bar that has to be met here.  Do these rules apply to all dashboards or is Home special in some way?

garath (approved):

missymessa:
> TIL and was surprised an accessibility compliance bar that has to be met here. Do these rules apply to all dashboards or is Home special in some way?

We get many a11y bugs for several areas of Grafana.

PR #6692, opened 2026-08-19
Use Status vault secrets for Grafana webhooks

missymessa:
> **Draft — do not merge in its current form.** Review feedback prompted an outcome audit of the V1 `.NET Status Alert` webhook. That audit shows the endpoint is producing noisy, conflated tracking rather than useful responder work.

## Outcome audit

- Since the Grafana webhook was changed to create AzDO work items, only one work item has ever been created with the `Grafana Alert` tag: [AB#11051](https://dev.azure.com/dnceng/internal/_workitems/edit/11051).
- The 43 successful `/api/alert` requests observed during the last 14 days did not create 43 actionable records. They repeatedly updated that same work item.
- AB#11051 now contains 107 comments: 104 from DotNetEng Status service identities, one automated triage comment, and two human comments.
- The automated comments conflate unrelated rules. An Android-device work item currently contains repeated `DotNetEng Status Failed Requests/Hour` notifications.
- The human investigation explicitly says those webhook comments are unrelated and should not be used as incident evidence.
- The actionable Android-device alert was independently routed to DDFun through Grafana-to-IcM and produced IcM 854214343, where an owner was assigned. The V1 AzDO webhook did not drive that response.

## Revised conclusion

The reviewer concern is valid: successful delivery does not demonstrate value. The current V1 webhook output is not cleanly actionable or meaningfully acted on, and preserving it does not justify adding generic cross-vault support to the Monitoring SDK.

This PR will not proceed with the current implementation. The appropriate direction is to retire the V1 `statusHook` routing/contact points and duplicated secrets, while the V2/IcM alert set continues side-by-side under AB#12093. The retirement edits need to ensure the legacy alert rules no longer route to this receiver; deleting only the source contact-point files would leave existing Grafana contact points behind.

AB#12004


garath (changes requested):
Marking as "request changes" just while I review this. I need to spend some time to understand what's going on.

src/Monitoring/Sdk/README.md
```diff
@@ -34,3 +34,9 @@ dotnet build MyMonitoring.proj `
 
 The deployment pipeline also supplies its Azure service-connection identity so the SDK can
 resolve Key Vault references.
+
+Use `[vault(secret-name)]` to resolve a secret from `GrafanaKeyVaultName`. To resolve a
```
garath: This seems to side-step a deliberate design decision. There are benefits to deferring the vault decision to a higher level.

garath:
Consider this pivot: I believe all notifications are currently disabled and the intent is to move them over to ICM. Does that mean we can deprecate this endpoint in dotneteng-status and remove these webhooks?

missymessa:
Good question. I checked the current configuration and production telemetry; we cannot safely remove this endpoint yet.

- The repository currently has 16 production alert rules routed to `.NET Status Alert`; only `cores-consumption` is paused, while the other 15 are configured as active.
- Production App Insights recorded 43 successful `POST /api/alert` requests (HTTP 204) during the last 14 days. The latest was 2026-08-20 14:52 UTC, so the endpoint is still actively receiving notifications.
- AB#12093 defines the V2/IcM migration as a side-by-side replacement and explicitly does not migrate the existing V1 alerts in place. Those V1 rules still depend on this webhook until each is retired or rerouted.

I agree that the end state should be to remove `/api/alert`, these contact points, and the duplicated secrets after the V1 rules have completed the IcM cutover. Until then, this PR is the bridge that keeps the active V1 path from drifting and prevents the AMG secret copy from breaking secret-manager synchronization.

garath:
> Production App Insights recorded 43 successful POST /api/alert requests (HTTP 204) during the last 14 days. The latest was 2026-08-20 14:52 UTC, so the endpoint is still actively receiving notifications.

Right, but are those notifications useful and acted on, or are they actually an oversight and the alerts should have been disabled?

missymessa:
You're right. I audited the outcomes rather than treating HTTP 204s as evidence of usefulness, and the V1 webhook is not producing actionable tracking.

- Only one `Grafana Alert` work item has ever been created since the AzDO migration: [AB#11051](https://dev.azure.com/dnceng/internal/_workitems/edit/11051).
- The 43 requests from the last 14 days repeatedly updated that same item rather than creating distinct actionable records.
- It now has 107 comments: 104 from DotNetEng Status service identities, one automated triage comment, and only two human comments.
- Unrelated rules are conflated: the Android-device item contains repeated `DotNetEng Status Failed Requests/Hour` notifications. The human investigation explicitly says those comments are unrelated and should not be used as incident evidence.
- The actual Android response was driven independently through Grafana-to-IcM (IcM 854214343), where DDFun received ownership; this webhook did not drive that action.

So yes: this is effectively an obsolete/noisy V1 route, not a useful bridge. I marked the PR draft and replaced the description with the evidence. I no longer think we should add cross-vault SDK support to preserve it. The revised direction is to remove the legacy `statusHook` routing/contact points and duplicate secrets, ensuring the rules are unrouted rather than merely deleting the source contact-point files.

missymessa:
Closing as superseded by the outcome-audit decision to retire the V1 statusHook path rather than preserve it with generic cross-vault support. The current implementation should not merge. Retirement work will be tracked under AB#12093 with explicit replacement-coverage, staged rollout, deployed-resource cleanup, and rollback criteria; AB#12004 will track removal of the duplicate AMG secrets after the V1 consumer is retired.

PR #6696, opened 2026-08-21
Add safe Grafana resource retirement support

missymessa:
## Context

This is prerequisite tooling for AB#12297, which will eventually retire the legacy V1 `.NET Status Alert*` routing and contact points from Azure Managed Grafana.

The Monitoring SDK currently creates or updates alert rules and contact points, but removing their source definitions does **not** remove the already-deployed Grafana resources. Without an explicit retirement path, old rules can continue firing and old contact points can remain live after their configuration is deleted from the repository.

The immediate V1 inventory contains 16 production rules, 15 of them active. Current Helix V2 definitions do not yet provide proven one-for-one coverage for those active rules, so the actual retirement remains blocked. This PR only adds a reusable, guarded mechanism: exact environment-specific manifests, report-only behavior by default, explicit delete opt-in, dependency ordering, reference checks, and deletion verification. It includes no retirement manifest and cannot delete anything as submitted.
## Summary

- add exact, environment-specific Grafana retirement plans for alert-rule UIDs and contact-point names
- keep retirement report-only unless `GrafanaAllowDeletes=true` is explicitly supplied
- delete alert rules before contact points, verify every deletion, and block contact-point deletion while a notification policy still references it
- delete every contact-point integration sharing the retired name

## Safety

This PR adds the mechanism only. It contains no retirement manifest, does not remove any alert definition or secret, and therefore performs no deletion when merged.

The follow-up V1 retirement change is blocked on AB#12297. Current source inventory proves 16 production V1 rules (15 active), while the current Helix V2 tree has no proven one-for-one replacement for those rules. The destructive manifest and source cleanup must not merge until replacement coverage or explicit retirement approval, staging validation, stakeholder sign-off, and rollback readiness are documented.

## Validation

- `dotnet test src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/Microsoft.DotNet.Monitoring.Sdk.Tests.csproj --configuration Release`

AB#12297


missymessa:
@garath @haruna99 gentle ping

garath:
I don't understand what this is for. I see the summary but I'm missing context.

missymessa:
Good question. This is prerequisite tooling for [AB#12297](https://dev.azure.com/dnceng/internal/_workitems/edit/12297), not the V1 retirement itself.

Today the Monitoring SDK creates/updates Grafana alert rules and contact points, but deleting their JSON/source definitions does **not** delete the deployed Grafana resources. That means a retired rule can keep firing and a legacy contact point can remain live unless someone deletes it manually.

The immediate use case is eventually retiring the V1 `.NET Status Alert*` routes and contact points. The current inventory has 16 production V1 rules (15 active), and we have not yet proven V2/IcM replacements for all of them, so the destructive retirement is intentionally blocked.

This PR adds the safe mechanism we will need once that gate is met:

- exact, environment-specific rule UIDs and contact-point names;
- report-only behavior unless `GrafanaAllowDeletes=true` is explicitly set;
- alert-rule deletion before contact-point deletion;
- refusal to delete a contact point still referenced by notification policy;
- idempotent deletion with verification.

There is no retirement manifest in this PR, so merging it cannot delete anything. A follow-up would supply the exact V1 resources only after replacement coverage or explicit retirement approval, staging validation, stakeholder sign-off, and rollback readiness.

haruna99 (approved):

garath (approved):

PR #6694, opened 2026-08-20
Add GitHub webhook secret drift alert

missymessa:
## Summary
- add production and staging alerts for DotNetEng Status GitHub webhook-secret drift
- detect sustained HTTP 400 failures on `/api/webhooks/incoming/github`
- fire when failures exceed 20 in the trailing hour for five minutes
- route to Teams independently of the failing GitHub webhook

The Grafana `/api/alert` HTTP 401 signal was removed after the V1 outcome audit. That endpoint, its contact points, and its duplicate secrets are being retired under AB#12297 and AB#12004 rather than preserved with new monitoring.

## Validation
- monitoring project builds with 0 warnings and 0 errors
- production incident-hour datasource backtest returned 364 GitHub webhook failures
- current production and staging hour returned no matching series
- the existing alert UID is preserved so deployment updates the staged rule rather than orphaning it

Tracking: https://dnceng.visualstudio.com/internal/_workitems/edit/11933
Related retirement: https://dev.azure.com/dnceng/internal/_workitems/edit/12297

## PR stack

Merge these PRs from top to bottom in this table. After a parent merges, retarget the next PR to `main`.

| Order | PR | Change |
| --- | --- | --- |
| 1 | [#6701](https://github.com/dotnet/dnceng/pull/6701) | Fix Grafana dashboard retirement cleanup |
| 2 | [#6703](https://github.com/dotnet/dnceng/pull/6703) | Migrate Data Migration alerts to Grafana |
| 3 | [#6694](https://github.com/dotnet/dnceng/pull/6694) | Add GitHub webhook secret drift alert |
| 4 | [#6702](https://github.com/dotnet/dnceng/pull/6702) | Retire source.dot.net monitoring |
| 5 | [#6706](https://github.com/dotnet/dnceng/pull/6706) | Inventory V1 Grafana retirements in report-only mode |

PR #6701, opened 2026-08-25
Fix Grafana dashboard retirement cleanup

missymessa:
## Context and deletion safety

The Monitoring SDK has long attempted to remove dashboards that it previously published but that are no longer present in source. It identifies those dashboards with ordinary Grafana dashboard tags written by the SDK:

- `baseuid:<uid>` records the exact source-derived dashboard UID.
- `source:<project>` records which monitoring project published it.

These are not Grafana-reserved tags. They are ownership markers defined by this SDK. Automatic stale-dashboard cleanup first searches for the exact `source:*` tag and then requires both that tag and a `baseuid:*` value matching the returned dashboard UID. A dashboard without both exact markers is excluded; arbitrary untagged or user-created dashboards are not swept.

The deployed dashboard with UID `home` is a one-time exception. It came from this repository's former `Monitoring.DncEng/dashboard/general/home.dashboard.json` (title `Home`), which PR #6679 removed. The deployed copy predates working ownership tags, so tag-guarded discovery cannot find it. This PR therefore names only the exact UID `home` in the staging and production retirement lists; it does not enumerate or delete other untagged dashboards. Grafana dashboard UIDs are unique within an organization, so the currently deployed `home` resource cannot simultaneously be a separate user dashboard.

The organization home preference is cleared before deletion because it currently points at UID `home`. Reversing that order would temporarily leave Grafana configured to open a missing dashboard. After staging and production both prove `/api/dashboards/uid/home` returns 404, the explicit `home` tombstone should be removed so a future user-created dashboard cannot inadvertently reuse an actively retired UID.
## Summary

- fix Monitoring SDK dashboard ownership tags so managed dashboards can be discovered and retired
- validate both the `baseuid:*` and `source:*` tags from Grafana search results before automatic deletion
- clear the Grafana home preference before deleting stale dashboards
- explicitly and idempotently retire the legacy untagged `home` dashboard in staging and production

## Root cause

PR #6679 removed the `home` dashboard definition and cleared the organization preference, but the existing cleanup path never discovered the deployed dashboard:

- ownership tags were added to one `JArray`, while a different array was serialized;
- the ownership guard looked for `baseuid:<uid>` as an object property instead of a member of the `tags` array.

Production currently has 123 dashboards and none carry the Monitoring SDK's `source:*` or `baseuid:*` tags. Successful deployments therefore cleared the preference but never attempted `Deleting extra dashboard home`.

The explicit retirement list handles that pre-existing untagged resource without treating arbitrary untagged dashboards as SDK-owned. Future deployments receive correct management tags and use guarded stale-resource cleanup.

## Validation

- `dotnet test src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/Microsoft.DotNet.Monitoring.Sdk.Tests.csproj --configuration Release`

Production verification after merge must confirm:

- `/api/dashboards/uid/home` returns 404;
- organization home-dashboard preference remains clear;
- both work-item waiting-time alert rules remain enabled and unlinked from dashboard panels.

Fixes AB#10233

## PR stack

Merge these PRs from top to bottom in this table. After a parent merges, retarget the next PR to `main`.

| Order | PR | Change |
| --- | --- | --- |
| 1 | [#6701](https://github.com/dotnet/dnceng/pull/6701) | Fix Grafana dashboard retirement cleanup |
| 2 | [#6703](https://github.com/dotnet/dnceng/pull/6703) | Migrate Data Migration alerts to Grafana |
| 3 | [#6694](https://github.com/dotnet/dnceng/pull/6694) | Add GitHub webhook secret drift alert |
| 4 | [#6702](https://github.com/dotnet/dnceng/pull/6702) | Retire source.dot.net monitoring |
| 5 | [#6706](https://github.com/dotnet/dnceng/pull/6706) | Inventory V1 Grafana retirements in report-only mode |

### DIFF
--- a/src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/GrafanaSerializationTests.cs
+++ b/src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/GrafanaSerializationTests.cs
@@ -204,4 +204,45 @@ internal class GrafanaSerializationTests
 
         act.Should().Throw<ArgumentException>();
     }
+
+    [Test]
+    public void SetDashboardManagementTags_ReplacesExistingManagementTags()
+    {
+        var dashboard = new JObject
+        {
+            ["tags"] = new JArray("custom", "baseuid:old", "source:old")
+        };
+
+        GrafanaSerialization.SetDashboardManagementTags(dashboard, "home", "source:Monitoring.DncEng");
+
+        dashboard["tags"].Values<string>().Should().BeEquivalentTo(
+            "custom",
+            "baseuid:home",
+            "source:Monitoring.DncEng");
+    }
+
+    [Test]
+    public void IsManagedDashboard_RequiresMatchingUidAndSourceTags()
+    {
+        var dashboard = new JObject
+        {
+            ["uid"] = "home",
+            ["tags"] = new JArray("baseuid:home", "source:Monitoring.DncEng")
+        };
+
+        GrafanaSerialization.IsManagedDashboard(dashboard, "source:Monitoring.DncEng").Should().BeTrue();
+
+        dashboard["tags"] = new JArray("baseuid:other", "source:Monitoring.DncEng");
+        GrafanaSerialization.IsManagedDashboard(dashboard, "source:Monitoring.DncEng").Should().BeFalse();
+
+        dashboard["tags"] = new JArray("baseuid:home", "source:other");
+        GrafanaSerialization.IsManagedDashboard(dashboard, "source:Monitoring.DncEng").Should().BeFalse();
+    }
+
+    [Test]
+    public void ParseDashboardUidList_TrimsAndDeduplicatesValues()
+    {
+        GrafanaSerialization.ParseDashboardUidList(" home,other;home ; ")
+            .Should().Equal("home", "other");
+    }
 }
--- a/src/Monitoring/Monitoring.DncEng/parameters.json
+++ b/src/Monitoring/Monitoring.DncEng/parameters.json
@@ -182,6 +182,13 @@
     }
   },
   {
+    "Name": "retired-dashboard-uids",
+    "Values": {
+      "Staging": "home",
+      "Production": "home"
+    }
+  },
+  {
     "Name": "dnceng-folderid",
     "Values": {
       "Staging": "46",
--- a/src/Monitoring/Sdk/DeployPublisher.cs
+++ b/src/Monitoring/Sdk/DeployPublisher.cs
@@ -82,9 +82,13 @@ public sealed class DeployPublisher : DeployToolBase, IDisposable
 
         await PostAlertRulesAsync().ConfigureAwait(false);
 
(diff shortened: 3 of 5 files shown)


garath:
I think I'm missing context. It reads like you're implementing some plan for dashboard management via tags but I'm missing details. 

>fix Monitoring SDK dashboard ownership tags so managed dashboards can be discovered and retired
 
What are "ownership tags"?

> validate both the baseuid:* and source:* tags from Grafana search results before automatic deletion

What are these tags? Are these Grafana things?

> clear the Grafana home preference before deleting stale dashboards

Why?

> explicitly and idempotently retire the legacy untagged home dashboard in staging and production

What is "legacy untagged" exactly? In particular, I want to be sure this won't delete any dashboards created by users.

missymessa:
Thanks; the missing distinction is between the normal tag-guarded cleanup and the one-time legacy `home` cleanup.

**What are the ownership tags?** They are ordinary Grafana dashboard tags written by our Monitoring SDK, not Grafana-reserved metadata:

- `baseuid:<uid>` identifies the exact source-derived UID.
- `source:<project>` identifies the monitoring project that published it.

The SDK searches by the exact `source:*` value and then requires **both** that tag and a `baseuid:*` value matching the dashboard's returned UID before automatic deletion. A dashboard without both exact markers is excluded. This is intended to prevent user-created or copied dashboards from being treated as SDK-owned merely because they are absent from this repository.

**Why clear the home preference first?** The Grafana organization preference currently points to UID `home`. We clear that preference first so Grafana falls back to its default homepage; deleting first would leave the organization configured to open a missing dashboard.

**What is “legacy untagged home”?** Until #6679, this repository published `Monitoring.DncEng/dashboard/general/home.dashboard.json`, with title `Home` and UID `home`. The deployed copy is that known SDK dashboard, but it predates working ownership tags—the bug fixed here wrote tags into an array that was never serialized. That is why normal tag-guarded cleanup cannot discover it.

The exception is deliberately narrow: `parameters.json` names only exact UID `home` for staging and production. It does not scan or delete other untagged dashboards, and Grafana UIDs are unique within an organization, so the currently deployed `home` resource cannot simultaneously be a different user dashboard.

There is one bounded caveat: while the explicit `home` tombstone remains configured, a future user-created dashboard deliberately reusing UID `home` would also be deleted. I updated the PR description to make the removal plan explicit: after staging and production both prove `/api/dashboards/uid/home` returns 404, remove the tombstone so that UID can be reused safely.

missymessa:
also: we deleted the home dashboard per a a11y issue, but it didn't get rolled out properly because of what this PR will fix. So I can't resolve the a11y issue yet.

garath:
I still don't understand why any of this is necessary. I see the _what_ just not the _why_. 

Grafana is a wild-west right now so I'm fine approving just about anything that doesn't violate security restrictions. As long as you have a plan, I'm good to go. I'm approving this PR under those considerations.

garath (approved):

PR #6702, opened 2026-08-26
Retire source.dot.net monitoring

missymessa:
## Summary

- remove the staging and production `source.dot.net` Grafana alert definitions
- remove the Source Browser availability panel and embedded legacy alert from the Service Availability dashboard
- retire the deployed `source-dot-net-availability` rules in both Grafana environments
- enable reviewed retirement plans during the managed Grafana deployment

The team no longer owns `source.dot.net`, so this PR removes its monitoring instead of changing the existing alert behavior.

## Validation

- `dotnet test src/Monitoring/Microsoft.DotNet.Monitoring.Sdk.Tests/Microsoft.DotNet.Monitoring.Sdk.Tests.csproj --configuration Release --no-restore` (12 passed)
- dashboard and retirement JSON parse successfully

Related: https://dev.azure.com/dnceng/internal/_workitems/edit/12284

## PR stack

Merge these PRs from top to bottom in this table. After a parent merges, retarget the next PR to `main`.

| Order | PR | Change |
| --- | --- | --- |
| 1 | [#6703](https://github.com/dotnet/dnceng/pull/6703) | Migrate Data Migration alerts to Grafana |
| 2 | [#6702](https://github.com/dotnet/dnceng/pull/6702) | Retire source.dot.net monitoring |
| 3 | [#6706](https://github.com/dotnet/dnceng/pull/6706) | Inventory V1 Grafana retirements in report-only mode |


MichaelSimons:
@missymessa, Friendly reminder: Please add one or two code reviewers PRs. If a PR isn't ready for review yet, please mark it as a draft until it is. Thanks!

missymessa:
@garath we can also remove this alert if we don't want it anymore :) (since we don't own this code anymore)

garath:
> @garath we can also remove this alert if we don't want it anymore :) (since we don't own this code anymore)

Two things: Yes, let's just remove any tracking of `source.dot.net`. Also, a lot of these changes look suspect. It's changing the alert behavior without any explanation why. What was wrong with the way it was configured? What does this new configuration bring?

garath (changes requested):
Holding due to comments in https://github.com/dotnet/dnceng/pull/6702#issuecomment-5430531232

missymessa:
Addressed in 1aaab94c. I removed the source.dot.net alert definitions from both environments rather than changing their behavior, removed the Source Browser panel and embedded legacy alert from the Service Availability dashboard, and added exact staging/production retirement plans so the deployed Grafana rules are deleted. The deployment now explicitly opts into reviewed retirement plans.

missymessa:
I also removed the hard-coded `GrafanaAllowDeletes=true` in d4c11577. The source.dot.net definitions and exact retirement entries remain, but the next staging deployment will be report-only so we can capture the proposed deletion before any destructive run. This matches the AB#12297 rollout gate. @garath, the behavior-changing version is gone and your direction to remove source.dot.net is now implemented non-destructively; please re-review.

missymessa:
To answer the configuration question directly: the superseded changes were intended to make the Grafana rule reproduce the Azure-native rule's behavior. Specifically, `mean` to `min` tested the worst location rather than averaging locations together; `<60` to `<61` represented the native `<=60` threshold using Grafana's strict comparison; removing the 15-minute `for` period and five-minute notification wait eliminated delays that were not present in the native rule; and the one-minute group interval matched the cadence observed on the deployed Grafana rule group.

That intent and evidence were not explained adequately in the original PR, and changing behavior is unnecessary now that we have confirmed the team no longer owns `source.dot.net`. Commit 1aaab94c removes those tuning changes along with the alert definitions and dashboard panel. The revised PR does not introduce new alert behavior; it retires the staging and production rules through the managed Grafana deletion path.

tl;dr: there was feedback you gave me on another PR to use Grafana for alerting instead of the native Azure alerts, so these changes are due to that request :)

garath (approved):

PR #6722, opened 2026-09-08
Allow Grafana deployment annotation queries

missymessa:
The Grafana deployment-annotations data source calls the dedicated endpoint without a GitHub user token, but the default API authorization convention currently rejects every request with 401. Production telemetry shows 2,756/2,756 requests failed over the last 30 days. Exempt only `GetGrafanaAnnotations` from the convention and cover that contract with a test. Relates to [AB#12448](https://dev.azure.com/dnceng/internal/_workitems/edit/12448).


MichaelSimons:
Friendly reminder to add reviewers.

garath (changes requested):
This seems like a step backwards. The API required authentication by design, and this undoes that without any real justification. 

It seems to me like the actual fix for the failures referenced in the description needs to happen in the Grafana plugin configuration.

missymessa:
Closing this approach based on the review feedback. The deployment-annotations Infinity datasource was intentionally provisioned with an Authorization header backed by the dotneteng-status-auth-header Key Vault secret, and the status API default policy deliberately restricts API access. The 401s show that the configured credential path is failing; they do not establish that Grafana cannot authenticate. AB#12448 will track investigation of the datasource credential (missing, expired, revoked, malformed, or incorrectly deployed) and a durable authenticated fix.

missymessa:
Replacement fix: https://github.com/dotnet/dnceng/pull/6725. It keeps the endpoint authenticated and configures the Infinity datasource to use standard Basic auth with the existing status API token.

